PR #42, opened 2013-03-14
OPML Support

stuporglue:
Good Morning!

I created an opml import page. You upload an opml XML file, and the entries are added to the database. 

I used it to import my Google Reader feeds into my own website and though others might want to do the same. 

Thanks,
Michael Moore

opml_1
opml_2

### DIFF
--- a/controllers/Opml.php
+++ b/controllers/Opml.php
@@ -1,0 +1,116 @@
+<?php
+/**
+ * @author Michael Moore <stuporglue@gmail.com>
+ * @license GPLv3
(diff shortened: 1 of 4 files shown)


jtojnar:
Hi @stuporglue. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions here and in other PRs under `GPL-3-or-later`?

PR #48, opened 2013-03-14
Switch from the shortcut.js to the jquery.hotkeys.js library.

Sigill:
The original shortcut.js library has trouble handling special keys.
For example, if a shortcut for 't' is defined, pressing F5 will
trigger it because the keycode for F5 is equal to the charcode for
't' (116).

See for example branch `throw_shortcuts_shortcut.js` in [git://github.com/Sigill/selfoss.git](git://github.com/Sigill/selfoss.git).

Switching to the jquery.hotkeys.js seemed to me a better option than patching the shortcut.js library.


jtojnar:
Hi @Sigill. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions here and in other PRs under `GPL-3-or-later`?

Sigill:
I agree with licensing my contributions to selfoss under GPL-3-or-later.

PR #62, opened 2013-03-15
Adding PostgreSQL support

volkadav:
Hello and thank you for making selfoss!  I noticed that the project doesn't presently support postgresql as a database, so I hacked up a patch to add that.  The versions I tested with were pg 9.0.7 and php 5.3.3-7 on debian stable i386; operationally I tested by importing my Google Reader OPML file and clicking around some.  The config.ini parameters are all the same, db_type would be set to pgsql (and port, likely, to 5432).  Probably any postgresql version 8.x or above would work (I checked the things I added against the 8.4 manual as I believe that is the version debian stable ships by default).

I'm guessing that anyone who wants to use this would know the relevant DBA tasks for creating a database and role, setting up authentication, etc.  There's a brief note in the header of the Database class pointing people at the online manual.

### DIFF
--- a/daos/pgsql/Items.php
+++ b/daos/pgsql/Items.php
@@ -152,7 +152,7 @@ class Items extends \daos\mysql\Items {
         if(is_numeric($sourceid)===false)
             return false;
         
-        $res = \F3::get('db')->exec('SELECT icon FROM items WHERE source=:sourceid AND icon!=0 ORDER BY ID DESC LIMIT 1',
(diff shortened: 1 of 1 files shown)


jtojnar:
Hi @volkadav. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions under `GPL-3-or-later`?

PR #68, opened 2013-03-15
Allow tag filtering to not include partial matches

GrumpyChunks:
When clicking on a tag (e.g. `foo`), that previously included partially matching tags (e.g. `football`). This fix corrects this.


jtojnar:
Hi @GrumpyChunks and @bsweeney. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions under `GPL-3-or-later`?

PR #75, opened 2013-03-16
Add an "open" button to the entry bar.

vincebusam:
Only for desktop mode.  The links on the icon and after the title were too small for my liking.  This could use an icon...

### DIFF
--- a/public/css/style.css
+++ b/public/css/style.css
@@ -737,6 +737,10 @@ body.publicmode.notloggedin .entry-unread {
             display:block;
         }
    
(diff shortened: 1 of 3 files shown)


jtojnar:
Hi @vincebusam. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions here and in other PRs under `GPL-3-or-later`?

PR #88, opened 2013-03-17
Make the home page configurable

ido50:
This change adds a new configuration option, "homepage", that allows selecting which page/filter/type to load by default. The default value for this option is "newest", which is selfoss's current behaviour, i.e. load the newest items by default. It can also take, however, the values "unread" or "starred", so the user can choose their preferred home page.

### DIFF
--- a/config.ini
+++ b/config.ini
@@ -18,3 +18,4 @@ rss_title=selfoss feed
 rss_max_items=300
 rss_mark_as_read=0
 readability_key=
+homepage=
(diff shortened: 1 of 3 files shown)


jtojnar:
Hi @ido50. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions here and in other PRs under `GPL-3-or-later`?

PR #90, opened 2013-03-17
Make home page configurable (second try)

jbfavre:
- Allow to define default homepage in config.ini (commit `cd55033b57`)
- Makes tags navigation respect current active nav-filter (commit `c272ed992c`)
- Makes hashChange from source respect current active nav-filter (commit `a938f1a24e`)

### DIFF
--- a/public/js/selfoss-events.js
+++ b/public/js/selfoss-events.js
@@ -38,7 +38,7 @@ selfoss.events = {
                 
             // from sources
             if(selfoss.events.lasthash=="#sources") {
-                $('#nav-filter li:first').click();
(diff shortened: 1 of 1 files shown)


jtojnar:
Hi @jbfavre. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions here and in other PRs under `GPL-3-or-later`?

jbfavre:
Hi @jtojnar 
I'm fine with `GPL-3-or-later` licensing.
Thanks for asking

PR #94, opened 2013-03-17
Share button

c4software:
Hi,

I have added a simple Google+ Share button into each item to simplify the share action of an article :

Share button

I have also inverted j/k and t/shift+t. I know (and understand) your position about the shortcut and the next item behavior.

If you don't want the j/k and next item placement modification just ignore modification in the "public/js/selfoss-shortcuts.js" file.

### DIFF
--- a/public/css/style.css
+++ b/public/css/style.css
@@ -447,6 +447,11 @@ input {
         .entry-toolbar .entry-sharetwitter {
             background-image:url("https://twitter.com/favicon.ico");
         }
+        
+        .entry-toolbar .entry-sharefacebook {
+            background-image:url("http://facebook.com/favicon.ico");
+        }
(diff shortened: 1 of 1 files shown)


jtojnar:
Hi @c4software. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions here and in other PRs under `GPL-3-or-later`?

-----

Also, for the reference, Tobias was referring to https://github.com/fossar/selfoss/commit/6eef0624273ea7108aee730fd103555f4959a379.

c4software:
Hi @jtojnar no problem for me.

PR #99, opened 2013-03-19
Displaying number of unread items for each tag

fedxa:
If you are interested, this displays and properly updates number of unread items for each tag.  However, the format is not very nice (especially in Desktop mode, when the tag label is small).

### DIFF
--- a/daos/mysql/Items.php
+++ b/daos/mysql/Items.php
@@ -367,27 +367,17 @@ class Items extends Database {
         return $res[0]['amount'];
     }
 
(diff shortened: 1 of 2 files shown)


jtojnar:
Hi @fedxa. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions here and in other PRs under `GPL-3-or-later`?

fedxa:
Hi -- do not really care.

PR #114, opened 2013-03-20
Adding unread counters for each tag

ghost:
I added an unread counter for each tag. It's compatible with multi-tagged items and with the "mark all read" button.

To test it, delete all.js file from public folder to regenerate it with the modified version of selfoss-events file.


jtojnar:
Looks like the pull request was #99.

PR #120, opened 2013-03-21
fix border around unread articles

pietheinstrengholt:
This stylesheet update fixes the border, at least in Google Chrome and probably other browser, for unread articles. I also added a better readability option for the span.unread. The span unread will also be hidden if no items are available.

### DIFF
--- a/daos/mysql/Items.php
+++ b/daos/mysql/Items.php
@@ -133,7 +133,7 @@ class Items extends Database {
      * @param string $uid
      */
     public function exists($uid) {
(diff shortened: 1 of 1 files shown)


jtojnar:
Hi @pietheinstrengholt. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions to https://github.com/fossar/selfoss/commit/4448b9d903aaaca8099793eb3a993a7b687b2927 under `GPL-3-or-later`?

PR #128, opened 2013-03-21
Add share to : http://getpocket.com , email

ploufs:
Add share item to pocket service (http://getpocket.com) and email.
Add target blank on selfoss link.


jtojnar:
Hi @ploufs. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions here and in other PRs under `GPL-3-or-later`?

PR #142, opened 2013-03-23
save the last update time of a source

lformella:
In my config there are to many sources and the update.php will always lead to a 500 server error, not being able to update the sources at the bottom. This change will also lead to 500 server errors, but will sequential update all source which are not up to date because they are sorted differently.


jtojnar:
Hi @lformella. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions here and in other PRs under `GPL-3-or-later`?

Issue #69, opened 2013-03-15
Folder Support 


koraa:
Hi,
I just tried to import an opml with folders and got the following debug message: "Debug opml import: invalid item $FOLER_NAME ".
I was able to work around this issue by removing the folder tags from the xml.

PR #158, opened 2013-03-25
Fix issue 69 : OPML with folders

ghost:
A little modification to handle OPML file with folders. #69

### DIFF
--- a/controllers/Opml.php
+++ b/controllers/Opml.php
@@ -88,16 +88,22 @@ class Opml extends BaseController {
         $errors = Array();
         
         // tags are the words of the outline parent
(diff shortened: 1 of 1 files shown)


jtojnar:
Hi @pablo-ruth. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions under `GPL-3-or-later`?

pablo-ruth:
Yes I'm fine with licensing my contributions under GPL-3-or-later.

PR #173, opened 2013-03-27
Turkish language file added

osmank3:


jtojnar:
Hi @osmank3. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions here and in other PRs under `GPL-3-or-later`?

osmank3:
> Hi @osmank3. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions here and in other PRs under `GPL-3-or-later`?
> 

No problem for me, you can license my contribitions gpl v3 or later

PR #191, opened 2013-03-28
Update on french langage (fr.ini)

GLLM:
2 addendum, a few changes & capital letters


jtojnar:
Hi @GLLM. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions here and in other PRs under `GPL-3-or-later`?

PR #214, opened 2013-04-02
Bold unread sources

crondog:
I have a few feeds which have very long names and so i cant tell if there are any unread items or not. Here are some changes which i feel would be beneficial to selfoss. I've never used jquery before so i hope this is ok :)


jtojnar:
Hi @crondog. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions under `GPL-3-or-later`?

crondog:
That is fine with me :+1:

PR #232, opened 2013-04-08
New Swedish language file

andersgd:
Language file for Swedish.


jtojnar:
Hi @andersgd. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions here and in other PRs under `GPL-3-or-later`?

andersgd:
Hi,

I’m fine with GPL-3-or-later 

Mvh
/Anders

PR #235, opened 2013-04-09
Fixed OPML import for other formats

RemyG:
Some OPML exports don't match the specifications. For example, [Netvibes](http://www.netvibes.com) exports OPML files without "type" or "text" attributes.
My changes allow the importer to still detect if an outline is a category or a feed entry, and it can still get the feed title even if there is no "text" attribute.

### DIFF
--- a/controllers/Opml.php
+++ b/controllers/Opml.php
@@ -99,9 +99,9 @@ class Opml extends BaseController {
         
         // parse every outline item
         foreach($xml->outline as $outline){
-            if((string)$outline['type']) {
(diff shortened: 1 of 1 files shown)


jtojnar:
Hi @RemyG. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions here and in other PRs under `GPL-3-or-later`?

RemyG:
Hi @jtojnar, I'm fine with this licensing setting.

PR #239, opened 2013-04-10
Create hu.ini

sanchomuzax:
The 1st version of Hungarian translation for Selfoss.


jtojnar:
Hi @sanchomuzax. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions here and in other PRs under `GPL-3-or-later`?

sanchomuzax:
Please, use it to free always. Its ready for me.

PR #244, opened 2013-04-11
Added autodiscovery to feed.php

rontu85:
I removed force_feed() from the spout feed.php and added set_autodiscovery_level() to automatically search for a feed in a given url.


jtojnar:
Hi @rontu85. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions here and in other PRs under `GPL-3-or-later`?

rontu85:
Sure, I'm fine with this.

PR #252, opened 2013-04-14
Plugin - Fetch Fulltext From Instapaper Parser

janeczku:
Get clean fulltext content for RSS items from Instapaper. Works much better than readability for many sources.

### DIFF
--- a/spouts/rss/instapaper.php
+++ b/spouts/rss/instapaper.php
@@ -1,0 +1,128 @@
+<?PHP 
+
+namespace spouts\rss;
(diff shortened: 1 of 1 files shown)


jtojnar:
Hi @janeczku. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions here and in other PRs under `GPL-3-or-later`?

Issue #242, opened 2013-04-11
[bug] Search function deletes from wrong tag


somebodygeeky:
If you search for multiple words (try "this is a test"), and try to delete "is", it will delete the "is" from "this", leaving the tags "th", "is", "a", "test".

PR #256, opened 2013-04-15
fix for #242

ochristi:
This should fix the error described in #242.
I added some code to id the parts and be able to delete the correct ones.
String.replace() used to remove the first occurrence of a search term and also left double spaces occasionally.


jtojnar:
Hi @ochristi. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions here and in other PRs under `GPL-3-or-later`?

PR #308, opened 2013-05-04
Sanitize tag list and title when adding/editing a source

seanrand:
I was getting slightly annoyed with having to use trim() and strlen() all over the place, so this cleans up the tag list into a nice, trimmed list when inserting or updating a source in the database.

Snippets like these:

```
foreach ($tags as $tag) {
    $tag = trim($tag);
     if (strlen($tag) > 0) {
         ...
```

would become redundant. (Well, except for tags that have been added to the database before this change, until the source is edited again...)

### DIFF
--- a/daos/mysql/Sources.php
+++ b/daos/mysql/Sources.php
@@ -22,9 +22,12 @@ class Sources extends Database {
      * @param mixed $params depends from spout
      */
     public function add($title, $tags, $spout, $params) {
+        // sanitize tag list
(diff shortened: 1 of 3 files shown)


jtojnar:
Hi @seanrand. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions here and in other PRs under `GPL-3-or-later`?

PR #329, opened 2013-05-10
updated fr.ini, french typo, etc.

gapato:
A little update to french translation to make it a little bit more french sounding.


jtojnar:
Hi @gapato. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions here and in other PRs under `GPL-3-or-later`?

PR #337, opened 2013-05-13
Optimise Expires section and Compression

sanand0:
I've merged the "Compression" and "Expires headers" sections from [html5boilerplate's .htaccess](https://github.com/h5bp/html5-boilerplate/blob/master/.htaccess) into selfoss/.htaccess

Compression was not working on the earlier version on a few servers, notably Hostgator's shared server setup. Web fonts also had a relatively short expiration time. The html5boilerplate code resolves both these issues.

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -37,45 +37,90 @@ RewriteRule (.*) public/$1 [L]
 
 # Default expires header if none specified (stay in browser cache for 7 days)
 <IfModule mod_expires.c>
-    ExpiresActive On
-    ExpiresDefault A604800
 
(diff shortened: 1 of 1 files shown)


jtojnar:
Hi @sanand0. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions under `GPL-3-or-later`?

sanand0:
Absolutely fine with `GPL-3-or-later`. Thanks!

PR #355, opened 2013-06-01
Update Authentication.php

zajad:
I've added a SSL proxy check. If a SSL proxy is used, domain and path of cookie have to be set differently.

### DIFF
--- a/helpers/Authentication.php
+++ b/helpers/Authentication.php
@@ -30,8 +30,16 @@ class Authentication {
      * start session and check login
      */
     public function __construct() {
+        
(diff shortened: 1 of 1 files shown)


jtojnar:
Hi @zajad. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions under `GPL-3-or-later`?

zajad:
Hi @jtojnar. All good and fine with 'GPL-3-or-later'.
Cheers

PR #357, opened 2013-06-04
Added methods to JSON API

Chumper:
I added two methods for the JSON API:
- /titles -> to get all items without the content, useful for mobile apps, that wants to display the title first and later load the entry.
- /items/@item -> to get a single item by id

I plan to build an Android app for the reader, this functions will be usefull so i don't have to load the whole feed at once.


amaccuish:
What happened to this? I'm building support for the Fever api and I need to be able to grab particular ids. Can't find this ability anywhere in the current selfoss api.

PR #365, opened 2013-06-10
Fix bug on Rss controller which not mark all item

titanz35:
When refreshing rss feed on client-app just one item (the last) is marked as read. The amrk as read instruction was written after the foreach loop, I re intregate it in the foreach.

### DIFF
--- a/controllers/Rss.php
+++ b/controllers/Rss.php
@@ -74,15 +74,16 @@ class Rss extends BaseController {
 
             $feedWriter->addItem($newItem);
             $lastid = $item['id'];
+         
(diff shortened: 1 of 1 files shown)


jtojnar:
Hi @titanz35. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions here and in other PRs under `GPL-3-or-later`?

titanz35:
Hi, no soucy ! BR

PR #372, opened 2013-06-21
add pro-linux spout

sgibb:
Add a spout for [pro-linux.de](http://www.pro-linux.de) (based on heise spout).

### DIFF
--- a/spouts/rss/pro-linux.php
+++ b/spouts/rss/pro-linux.php
@@ -1,0 +1,196 @@
+<?PHP
+
+namespace spouts\rss;
+
+/**
+ * Plugin for fetching the news from pro-linux with the full text.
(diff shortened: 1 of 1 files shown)


jtojnar:
Hi @sgibb. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions under `GPL-3-or-later`?

sgibb:
I am fine with `GPL-3-or-later`. On the other side this contribution should be deprecated anyway because www.pro-linux.de is a read-only site without any updates since May 2020. (And they clearly wrote that they switch to read-only and won't update their site.)

PR #373, opened 2013-06-23
 Add Romansh (rm) translation 

gion-andri:
Adding Romansh (rm) translation

### DIFF
--- a/public/lang/rm.ini
+++ b/public/lang/rm.ini
@@ -1,6 +1,6 @@
-lang_markread=mark as read
-lang_newest=newest
-lang_unread=unread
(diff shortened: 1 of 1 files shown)


jtojnar:
Hi @gion-andri. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions here and in other PRs under `GPL-3-or-later`?

gion-andri:
@jtojnar Of course. Its ok, if my contributions are licenced under GPL-3-or-later. 
Thank you!

PR #374, opened 2013-06-26
Adds support for PHP 5.4+ embeddable web server 

dz0ny:
User simply runs with php -S 0.0.0.0:8000 run.php and app works.
No need for other servers, this is super useful for RPI or testing, developing.

PHP 5.4+ only


jtojnar:
Replaced by https://github.com/fossar/selfoss/pull/376

PR #381, opened 2013-07-02
Shorten german translation for mark and unmark

snipem:
The german translation for mark and unmark is shortened in order to make picture download possible on mobile browsers like iOS Safari


jtojnar:
Hi @snipem. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions under `GPL-3-or-later`?

PR #382, opened 2013-07-02
Add a finnish translation

jukper:
I created a finnish translation for selfoss.


jtojnar:
Hi @jukper. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contribution under `GPL-3-or-later`?

PR #387, opened 2013-07-11
Fix for #386

### LINKED ISSUE #386: No Open link in toolbar visible at iPad
Hi, I can not see the Open link (entry-newwindow) when using selfoss at the iPad, see screenshot. The LI element seems to be there though. On iPhone and  Desktop it is visible as expected. I'm using selfoss 2.8, latest Git checkout.

no-newwindow-link

Weishaupt:
This commit fixes #386.

One should ponder, if the smartphone view should be different in terms of "open in a new window" from all other form factors.

### DIFF
--- a/public/css/style.css
+++ b/public/css/style.css
@@ -969,10 +969,6 @@ body.publicmode.notloggedin .entry-unread {
             display:block;
         }
    
-    .entry-toolbar .entry-newwindow {
-        display:none;
-    }
-
     .entry-content {
         display:none;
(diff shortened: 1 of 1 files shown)


jtojnar:
Hi @Weishaupt. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions here and in other PRs under `GPL-3-or-later`?

PR #390, opened 2013-07-23
Allow entry share to be configured

becevka:
For example, in config.ini set share=frde. 
That will show only facebook, readability, delicious and email share buttons. 
Respects order.

### DIFF
--- a/public/css/style.css
+++ b/public/css/style.css
@@ -262,10 +262,16 @@ input {
         color:#555;
         text-decoration:none;
     }
- 
-.slider {
+    
+
+/* search */
+
+#search {
     z-index:9;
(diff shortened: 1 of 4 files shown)


jtojnar:
Hi @becevka. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions here and in other PRs under `GPL-3-or-later`?

becevka:
Hi @jtojnar, thanks for reaching out.
I am fine to license my contributions under GPL3+ license or any other license selfoss needs to support.

Issue #105, opened 2013-03-19
mysql prefix


bhopmann:
It would be better to have the opportunity of setting a sql table prefix. So there won't be interferences with any identically named tables.

PR #391, opened 2013-07-23
Add table prefix parameter for MySQL engine

mistic100:
following this feature request
https://github.com/SSilence/selfoss/issues/105


jtojnar:
Hi @mistic100. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions here and in other PRs under `GPL-3-or-later`?

mistic100:
@jtojnar No problem

PR #400, opened 2013-07-31
Enter missing values

matmattia:
I enter "lang_source_export", "lang_login", "lang_login_username", "lang_login_password" e "lang_no_title".


jtojnar:
Hi @matmattia. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contribution under `GPL-3-or-later`?

PR #406, opened 2013-08-08
Fixes #201 and adds a global error bar at the top 

### LINKED ISSUE #201: Do not display a javascript error popup if no network when at the bottom of the list of articles
Hi all,

if you loose your network connectivity and reach to bottom of articles list, you get a Javascript error popup several times !

A bottom banner (like an article but with a different background color/font/anything) indicating the problem would be best wouldn't it ? Instead of the popup

tamizhgeek:


jtojnar:
Hi @tamizhgeek. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions here and in other PRs under `GPL-3-or-later`?

tamizhgeek:
Hello Jan,

Yes I am fine with that.

Issue #409, opened 2013-08-15
Selfoss launches letterbox from iPhone 5 Springboard (possible fix included)


bjoerns1983:
I have seen that Selfoss, if added to the Homescreen of an iPhone 5 only loads in letterbox mode, not using the full screen size.

This behaviour can be changed by adding this
`<meta name="viewport" content="initial-scale=1.0,user-scalable=no,maximum-scale=1" media="(device-height: 568px)" />`
to templates/home.phtml and other which hold the viewport meta.

I didnt test that for sidefeects to other devices, but on my iPhone 5 selfoss now start normal from the Springboard.

PR #410, opened 2013-08-16
Add support for fullscreen Webapp on iPhone 5

bjoerns1983:
Refers to Issue #409 where the Webapp on an iPhone 5 started from
Springboard starts letterboxed.


jtojnar:
Hi @bjoerns1983. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions under `GPL-3-or-later`?

PR #415, opened 2013-08-23
Fix 'Undefined Index' error in item tpl when no shares available

vntw:
There´s nothing in `$share` if I didn´t choose any shares - result: an 'Undefined Index' error will be raised.

Fix: Check if any shares are available


jtojnar:
Hi @venyii. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions under `GPL-3-or-later`?

venyii:
Hey, yes I'm fine with that 👍

PR #419, opened 2013-08-28
Wider touch area

silentboatman:
Wider touch area for the settings icon, in mobile view.

Touch area before:
before

Touch area after:
after

### DIFF
--- a/public/css/style.css
+++ b/public/css/style.css
@@ -1051,10 +1051,10 @@ body.publicmode.notloggedin .entry-unread {
             width:29px;
             height:29px;
             left:100%;
-            top:10px;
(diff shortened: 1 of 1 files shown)


jtojnar:
Hi @silentboatman. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions under `GPL-3-or-later`?

silentboatman:
Hi! Licensing my contributions to selfoss under GNU GPL 3 or any later version is absolutely fine.

PR #438, opened 2013-10-03
Improved performance by reducing the number of mysql queries.

koehn:
This same change will need to be added to the other Items daos, but this took my number of mysql queries down from over 20,000 to about 100.

### DIFF
--- a/daos/mysql/Items.php
+++ b/daos/mysql/Items.php
@@ -87,10 +87,6 @@ class Items extends Database {
      * @param mixed $values
      */
     public function add($values) {
-        // don't add items twice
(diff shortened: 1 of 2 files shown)


jtojnar:
Hi @koehn. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contribution under `GPL-3-or-later`?

koehn:
Wow that's from a long time ago! I'm fine licensing my contribution under `GPL-3-or-later`.

PR #442, opened 2013-10-18
Add more robust RSS feed date parsing

rnavarro:
Added the ability to specify different formats for parsing RSS feed publish dates. Some of the RSS feeds have screwy formats.

This should be backwards compatible with the existing parser....I tested it and it seemed to work perfectly.


jtojnar:
Hi @rnavarro. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions here and in other PRs under `GPL-3-or-later`?

PR #444, opened 2013-10-31
reddit2 spout: add empty validation on username&password to fix unable t...

bdeferme:
I was unable to add a reddit2 source, the error I got was "Undefined index: validation".
These changes (empty validation because the fields are not required) fixes that.


jtojnar:
Hi @bdeferme. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions here and in other PRs under `GPL-3-or-later`?

PR #458, opened 2013-11-26
Update home.phtml

VincentCasse:
Delete example error message after darkside commit


jtojnar:
Hi @VincentCasse. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions under `GPL-3-or-later`?

PR #463, opened 2013-12-14
Heise

robelix:
heise spout finetuning:
- remove javascript from content
- new text deliminator for for some tp articles


jtojnar:
Hi @robelix. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions here and in other PRs under `GPL-3-or-later`?

robelix:
GPL3-or-later is ok for me.

Sorry for late reply.

PR #465, opened 2013-12-20
Setting to lazy load images on mobile devices

bodnaristvan:
Added new setting called 'load_images_on_mobile', if set to 1 in config.ini images will be autoloaded on mobile devices.


jtojnar:
Hi @bodnaristvan. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions under `GPL-3-or-later`?

bodnaristvan:
@jtojnar yeah, i'm fine with that! thanks for the heads up

PR #497, opened 2014-03-07
Hide sources and tags with no unread articles

mrahmadt:
Small change in the code to hide sources and tags with no unread articles

### DIFF
--- a/templates/item.phtml
+++ b/templates/item.phtml
@@ -36,13 +36,13 @@
     
     <!-- title -->
     <h2 class="entry-title"><?PHP echo $title; ?></h2>
-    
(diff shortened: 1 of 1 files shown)


jtojnar:
I have pushed a [patch](https://github.com/fossar/selfoss/commit/0bb38e08d5e5fe87b4d996f44f37e6deb9a9fe5c) that adds a class to all tags and sources in the sidebar that have no unread items, allowing you to hide them by adding the following CSS to the `user.css` file in your selfoss directory:

```css
#nav-tags li.read, #nav-sources li.read {
  display: none;
}
```

Hiding tags in the main content is similarly possible using:

```css
.entry-tags {
  display: none;
}
```

PR #510, opened 2014-04-08
Little fix to Polish translation

Siarkowy:
Use proper grammar.


jtojnar:
Hi @Siarkowy. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions under `GPL-3-or-later`?

Siarkowy:
No problem, @jtojnar!

PR #523, opened 2014-05-13
added estonian translation

kristiank:


jtojnar:
Hi @kristiank. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contribution under `GPL-3-or-later`?

kristiank:
Hi @jtojnar. Thank you for asking, I am very fine with licensing my contributions under `GPL-3-or-later`.

PR #524, opened 2014-05-16
allow <sub> and <sup> elements

hdhoang:
some articles mention "2<sup>60</sup> bits" and come out as "260 bits"


jtojnar:
Hi @hdhoang. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions here and in other PRs under `GPL-3-or-later`?

hdhoang:
I'm fine with licensing my contributions under GPL-3-or-later, thanks for getting back.

PR #532, opened 2014-07-04
REST API : Get only items updated since given time

yDelouis:
Hi,

First of all, thanks for your great work.

I'm developing an Android client for Selfoss and I'm facing an issue for syncing articles.
At the moment, there is no way to know which items have been modified (new or marked read) since the last sync. So the client can't stay in sync with the server without downloading all articles at each sync.

This PR enables the server to track update of articles and add a REST parameter for items to get only the articles updated since a given timestamp. With this pieces of code, the client would be able to keep in sync efficiently with the server.

I have tested both upgrading and new install for each 3 database (sqlite, mysql and postgresql).

Let me know if there is something wrong.

### DIFF
--- a/daos/mysql/Items.php
+++ b/daos/mysql/Items.php
@@ -220,7 +220,13 @@ class Items extends Database {
             $params[':source'] = array($options['source'], \PDO::PARAM_INT);
             $where .= " AND items.source=:source ";
         }
-        
+
+        // update time filter
+        if(isset($options['updatedsince']) && strlen($options['updatedsince'])>0) {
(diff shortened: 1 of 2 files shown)


jtojnar:
Hi @yDelouis. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions here and in other PRs under `GPL-3-or-later`?

yDelouis:
@jtojnar Yes ! I'm fine with this !

PR #538, opened 2014-07-17
Add wallabag support

alistair-marshall:
Add wallabag (https://www.wallabag.org) support to sharing entries. Wallabag is an open source, self hosted read it later type of application.

### DIFF
--- a/_docs/website/index.html
+++ b/_docs/website/index.html
@@ -284,7 +284,11 @@ db_port=3306</code>
                     </tr>
                     <tr>
                         <td class="documentation-first-column">share</td>
-                        <td>share defines which share buttons beneath the entry are visible. Default is share=gtfprde (g = google, f = facebook, t = twitter, p = pocket, r = readability, d = delicious, e = email). When you would like to show only facebook and twitter share button use share=ft.</td>
+                        <td>share defines which share buttons beneath the entry are visible. Default is share=gtfprde (g = google, f = facebook, t = twitter, p = pocket, r = readability, d = delicious, w = wallabag, e = email). When you would like to show only facebook and twitter share button use share=ft.</td>
+                    </tr>
+                    <tr>
+                        <td class="documentation-first-column">wallabag</td>
+                        <td>wallabag url. url to wallabag homepage</td>
                     </tr>
                     <tr>
                         <td class="documentation-first-column">unread_order</td>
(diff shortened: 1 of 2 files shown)


FromBabylon:
I have put my wallabag server adress but I do not have any wallabag button for sharing. I have all other buttons like Google, readability etc.. but no wallabag one

jtojnar:
@FromBabylon have you also added `w` into share property in your config? See [documentation](https://github.com/SSilence/selfoss/pull/538/files#diff-d0d23579773acbdc9b3e380ba104879bR287)?

FromBabylon:
I forgot ! 
Thanks it works now

Poorchop:
I had this working with my wallabag v1 installation but since upgrading to v2, this no longer seems to work. I changed some of the potentially personal information but this is the error:

```
2016/05/23 01:21:20 [error] 1963#0: *48 FastCGI sent in stderr: "PHP message: HTTP 404 (GET /wallabag.example.com/)" while reading response header from upstream, client: 11.22.33.44, server: selfoss.example.com, request: "GET /wallabag.example.com/?action=add&url=SomeString= HTTP/1.1", upstream: "fastcgi://127.0.0.1:9000", host: "selfoss.example.com"
```

tcitworld:
Hi,
The bookmarklet doesn't work - yet - in some cases with v2. Please use the v2 browser plugin if you use Firefox.

Poorchop:
Okay, thanks. Is selfoss+wallabag integration going to happen again in the future?

tcitworld:
We at wallabag would certainly like too. We've got a brand [new API](http://v2.wallabag.org/api/doc). :-)

jtojnar:
Hi @alistair-marshall. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions under `GPL-3-or-later`?

alistair-marshall:
> Hi @alistair-marshall. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions under `GPL-3-or-later`?
> 

No problem with V3 or later.

PR #544, opened 2014-08-14
Add Twitter media and overhaul displaying tweets

Alcatr4zz:
The Twitter spout is now much more usable. Tweeted photos are directly integrated within the stream as thumbnails, tweets are better readable, links are clickable and retweets are better recognizable as such.

twitter

Technical:
The content for the Twitter spout has been removed. Titles are parsed as html enabling the use of line breaks and links.
Thumbnails have been resized for better optics. Their size (especially on mobile devices) is limited to the screen width.

### DIFF
--- a/public/css/style.css
+++ b/public/css/style.css
@@ -426,6 +426,10 @@ input {
         padding-bottom:7px;
     }
     
+        .entry-title a {
+            color:#3d6d69;
+        }
+    
         .entry.unread  .entry-title {
             color:#333333;
         }


jtojnar:
Hi @Alcatr4zz. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions here and in other PRs under `GPL-3-or-later`?

Alcatr4zz:
Hi @jtojnar I'm fine with licensing all my contributions under GPL-3-or-later.

PR #548, opened 2014-09-03
Show unread count in the document title

MTGap:
Shows the number of unread items in the document title and updates the text as read statuses change.


jtojnar:
Hi @MTGap. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions here and in other PRs under `GPL-3-or-later`?

PR #556, opened 2014-09-18
Update Polish translation

mg6:
Fix for Polish translation. This could be better but as of my knowledge, F3 framework does not allow the translator to specify correct format string for Polish language at the moment.


jtojnar:
Hi @mg6. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions here and in other PRs under `GPL-3-or-later`?

mg6:
Sure, I'm fine with licensing under “GPL v3 or later” for all my contributions to date.

PR #573, opened 2014-09-30
Throw button on toolbar

mkholt:
I have often missed being able to "throw" an entry on the mobile interface.
On desktop, it's possible to push the "t" button to mark the current entry as read and display the next one.
This is not practical on mobile, so I have instead added a "Throw" button to the interface.
On desktop, this is placed in the toolbar with the other buttons, and on mobile it is placed in the bottom toolbar along with the "Open".
It is placed in the bottom, since it is most likely to be used after reading an entry. (At least that is how I use it).

Note, it is not possible to use the "nextprev" shortcut function, as this does not behave as expected in the mobile interface.

### DIFF
--- a/public/js/selfoss-events-entriestoolbar.js
+++ b/public/js/selfoss-events-entriestoolbar.js
@@ -28,10 +28,11 @@ selfoss.events.entriesToolbar = function(parent) {
 
     // throw item
     parent.find('.entry-throw').unbind('click').click(function(e) {
-        var $selected = $('.entry.selected');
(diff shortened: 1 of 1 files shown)


jtojnar:
Hi @mkholt. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions here and in other PRs under `GPL-3-or-later`?

mkholt:
I have no issue with my contributions being licensed under `GPL-3-or-later`

PR #574, opened 2014-09-30
Prevents an error fetching favicons.

chrisguidry:
When encountering a favicon `<link>` where the `href` is not enclosed
in quotes, the selfoss `ContentLoader` would fail and exit early.

For example, `<link rel="shortcut icon" href=img/favicon.ico>` would cause
the error `Undefined offset 2` in `helpers/Image.php`.  Making the regex
a bit more liberal and also checking the results of the final `preg_match`
fix this.

This fixes the issue I asked about in http://selfoss.aditu.de/forum/index.php?id=936.
This is also the corrected followup to the incorrectly-based PR in https://github.com/SSilence/selfoss/pull/572.

### DIFF
--- a/helpers/Image.php
+++ b/helpers/Image.php
@@ -162,8 +162,9 @@ class Image {
         if($result==0)
             $result = preg_match('/<link [^>]*rel=("|\')icon\1.*>/Ui', $html, $match1);
         if($result>0) {
-            preg_match('/href=("|\')(.*)\1/Ui', $match1[0], $match2);
-            return $match2[2];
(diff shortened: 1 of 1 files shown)


jtojnar:
Hi @chrisguidry. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions here and in other PRs under `GPL-3-or-later`?

PR #585, opened 2014-10-09
When front-end nginx (through https:443) forward to apache2 (through htt...

auzias:
According to some use-cases with apache2/nginx proposes client resource through http instead of https. This includes all `href` and `src` ressources. Web browser, by default, block (security reasons) theses ressources.

As [asked](https://github.com/SSilence/selfoss/pull/582#issuecomment-58396733) I put the changes (lighter btw) at the right place.

Note that nginx front-end users need to location as this:

```
      location / {
         proxy_set_header X-Nginx-Scheme $scheme;
          proxy_set_header X-Forwarded-Port $server_port;
          proxy_pass   http://<whatever IP you need>:<whatever port you need>/;
          include      /etc/nginx/proxy.conf;
      }
```

I think it good to pull now :) :bug: (otherwise, just ask)

### DIFF
--- a/helpers/View.php
+++ b/helpers/View.php
@@ -48,14 +48,19 @@ class View {
             $subdir = $lastSlash!==false ? substr($_SERVER['SCRIPT_NAME'], 0, $lastSlash) : '';
             
             $protocol = 'http';
-            if (isset($_SERVER["HTTPS"]) && ($_SERVER["HTTPS"]=="on" || $_SERVER["HTTPS"]==1))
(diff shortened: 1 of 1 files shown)


jtojnar:
Hi @auzias. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions here and in other PRs under `GPL-3-or-later`?

auzias:
I do !
The most permissive license is OK for me.

PR #590, opened 2014-10-20
added an option to not delete the unread items

janw-oostendorp:
Allow to preserve unread item's that are over the 'duedate'  
Preserves the original behaviour in the `default.ini` or if the setting is missing


SSilence:
Thanks a lot for the pr but I don't think this option is really necessary. 

Thanks again
 Tobi

PR #591, opened 2014-10-21
Various enhancements

niol:
This pull request contains two small enhancements and a modification of how permission is denied to the application when not logged in: now the application returns a 403 error. This makes it possible to handle this particular error in js, e.g. presenting a logging prompt when an AJAX call fails because the session has expired.

It also adds a way to get a user agent for spouts.

It also fixes some stuff in getting feed icons.

(details in the commits)

### DIFF
--- a/controllers/BaseController.php
+++ b/controllers/BaseController.php
@@ -28,4 +28,26 @@ class BaseController {
     public function __construct() {
         $this->view = new \helpers\View();
     }
+
+    /**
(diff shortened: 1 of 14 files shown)


niol:
Is there any problem preventing accepting this PR?

SSilence:
There is no problem, I'm only lack of time and will check this pr the next days. Thanks a lot for this pr!!!

PR #597, opened 2014-11-07
Some enhancements on source updating

niol:


niol:
Is there any problem preventing accepting this PR?

SSilence:
I don't think this is really necessary. It should be the normal usecase that a cronjob updates only 20-30 Minutes. Thanks a lot for this and the many other pull requests!!!

niol:
There were many bug reports related to this. But like you, this is not very useful for my use case.

niol:
If you use the new sync button while the cron script is running, you may get duplicate items
In the db. This PR prevents this.

jcbrand:
> If you use the new sync button while the cron script is running, you may get duplicate items
In the db. This PR prevents this.

That sounds useful.

Issue #592, opened 2014-10-27
Filter in the text?


debby1:
Hi, I would like to know if on selfoss there is the feature filter in the text.

I need a Rss reader on server to receive many feeds from Facebook pages of companies and site webs, but only about some and specific things.

For example: I need to know news only about "recipes", so I would like to read only posts from Facebook pages with the word "recipe" in the text, and not all rest of posts..

Is this possibile? Thank you very much in advance

niol:
This can be closed: fixed in 2.13.

Issue #449, opened 2013-11-04
Still no filter?


Koopzington:
I just wanted to know if you still don't have the intention to add a filter functionality like mentioned in #15 since i'd really appreciate such a thing.
Everything i'd personally need would be just a blacklist of words which shouldn't show up in the titles and therefore shouldn't be added to the unreaded items list...

But i guess you'd also need to add a whitelist and both configurable for global AND for each single source once you'd have started implementing filters...

niol:
This has been implemented in 2.13.

PR #606, opened 2014-12-24
Add filter feature

especially-relative:
This PR adds the filter feature from rsslounge to selfoss. Addresses open issues #592 and #449.


jtojnar:
Hi @danieljkemp. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions under `GPL-3-or-later`?

PR #610, opened 2015-01-07
Feat new helper urlcleaner

jbfavre:
The purpose of this helper is to perform curl calls to get real URL.
When get, it also try to remove known trackers which are integrated as URL params or fragment.

### DIFF
--- a/helpers/UrlCleaner.php
+++ b/helpers/UrlCleaner.php
@@ -4,7 +4,8 @@ namespace helpers;
 
 /**
  * Helper class for getting real URL from feed
- * as well as removing tracking params
+ * as well as removing tracking params from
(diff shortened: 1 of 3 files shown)


niol:
Using curl helps workaround ```file_get_contents()``` bugs but depends on ```php-curl```. Maybe it is better to use ```f3``` internals for this, like I propose in #591 and f0f609c723d44f74f6e2f51df02317d8b046c533 .

Can you specify what removing redirects is trying to fix?

jbfavre:
@niol providing real URL instead of feed one (for example feedburner) makes it easier to share articles from selfoss.
Same goes for removing tracker, adding privacy enhancements

PR #611, opened 2015-01-09
Add session keep alive

PhrozenByte:
Send periodic AJAX requests to keep the session alive. About every 10 minutes a request is sent, which actually does nothing, but prevents PHP from garbage collecting the session file. When the user is neither logged in nor Selfoss is in public mode, the users session expired and can't be re-initialized automatically, as it would be possible in public mode. If this is the case, reload the page, which will present the login form.

### DIFF
--- a/public/js/selfoss-base.js
+++ b/public/js/selfoss-base.js
@@ -20,11 +20,16 @@ var selfoss = {
         source: '',
         ajax: true
     },
-
+    
     /**
      * instance of the currently running XHR that is used to reload the items list
      */
     activeAjaxReq: null,
+    
+    /**
+     * interval ID of session keep alive
+     */
+    keepAliveIntervalID: null,
     
     /**
      * initialize application
@@ -36,7 +41,10 @@ var selfoss = {
                 $('#username').focus();
                 return;
             }
-        
+            
+            // init session keep alive
+            selfoss.keepAliveIntervalID = setInterval(selfoss.keepAlive, 590000);
+            
             // set items per page
             selfoss.filter.itemsPerPage = $('#config').data('items_perpage');
             
@@ -48,6 +56,30 @@ var selfoss = {
             
             // init shortcut handler
             selfoss.shortcuts.init();
+        });
+    },
+    
+    
+    /**
+     * keep session alive by sending periodic ajax requests
+     *
+     * @return void
+     */
+    keepAlive: function() {
+        $.ajax({
+            url: $('base').attr('href') + 'keepAlive',
+            type: 'GET',
+            dataType: 'json',
+            success: function(data, textStatus, jqXHR) {
+                // neither logged in nor in public mode
+                // redirect to login page
+                if (!data.publicMode && !data.loggedin) {
+                    location.reload();
+                }
+            },
+            error: function(jqXHR, textStatus, errorThrown) {
+                clearInterval(selfoss.keepAliveIntervalID);
+            }
         });
     },
     
--- a/controllers/Index.php
+++ b/controllers/Index.php
@@ -170,6 +170,20 @@ class Index extends BaseController {
     
     
     /**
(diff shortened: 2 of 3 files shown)


public/js/selfoss-base.js
```diff
@@ -20,13 +20,18 @@ var selfoss = {
         source: '',
         ajax: true
     },
-
+
```
jtojnar: superfluous whitespace

jtojnar:
What if you are reading a really long article in selfoss? When the page gets reloaded in the middle you will probably get angry.

public/js/selfoss-base.js
PhrozenByte: You're telling me that despite the fact that Selfoss has no coding guidelines and this is inconsistent over the whole codebase. I think SSilence focuses on more important things ;)

PhrozenByte:
The point about this pull request is that your session never expires unless you close your browser window. So, which circumstances have you in mind?

jtojnar:
1. You open a really long article
2. In new tab you open some definition
3. All timers are paused by brows
3. Ten minutes are over but

jtojnar:
1. You open a really long article
2. In new tab you open some definition
3. All timers are paused by browser
4. Ten minutes are over and you are logged off
5. You switch back to previous tab, timers are unfreezed callback gets called and the page is reloaded
6. You lose track of where you were in the article

public/js/selfoss-base.js
jtojnar: I understand the coding isn’t really that important but let’s not make it worse please.

PhrozenByte: What gives you the idea, that there should be no whitespaces? Do you really want to discuss such trivia? I don't want to. So, anyway...

PhrozenByte:
Why exactly should a browser freeze this timer? Never heared about that. The only thing I've heared is that browsers delay timers with execution times less than 1000 ms on inactive tabs.

Anyway, lets see what SSilence says.

public/js/selfoss-base.js
jtojnar: Whitespace is used for indentation, what is it for when there is nothing to indent? OK, I’ll leave it at this and make a pull request cleaning the coding style of the whole project later.

jtojnar:
There is **no guarantee** it won’t. There was this [extension](https://adblockplus.org/blog/preventing-background-tabs-from-wasting-your-computer-s-resources) once. Another possible cause is your computer automatically going to hibernation because of low battery level. Initially I was worried about loss of network connection but you have that covered by removing the timer on failure.

I realize these are edge cases but I still think asking user for confirmation before reload would be much safer.

PhrozenByte:
JavaScript timers are clock-based, you know? :/
When your computer goes to hibernation for longer than 10 minutes, the timer is executed immediately after wakeup. This all has been reconsidered already...

jtojnar:
And when the timer is executed session has already expired so the app is reloaded.

PhrozenByte:
Works as desired. Never leave the user in a unusable environment.
As said, neither you, nor me takes the decision, it's up to SSilence what he wants in **his** project. But don't talk down my code wrongful, thats rude. I have no more to say.

jtojnar:
I am not trying to talk down your work wrongful (I am not a native English speaker so I am sorry if it came out that way), I only try to address possible problems that could potentially disturb UX. The environment is AFAIK usable if you’ve got open a long article and you are signed out (due to one of things mentioned before), you can still continue reading it. If you finish reading it and carry out some action selfoss itself will show you login form. But it won’t try to reload the page while you are (seemingly) inactive.

Keeping the session alive is great and you’ve done a good work, but I would remove the reload, or at least wrap it in a prompt to prevent the aforementioned issue.

SSilence:
Thanks a lot for the pr. I think session timeout is a server configuration thing.

PR #617, opened 2015-01-25
Do the AddHandler php5-cgi only if the server is not running mod_php

robelix:


wAmpIre:
IMHO the handler for php SHOULD (MUST) NOT be changed by .htaccess. This a server/admin thing, not software/user

robelix:
I did not add the handler, I just changed it to not use it if mod_php is running.
And it's fine for me if it's removed completely.

SSilence:
I agree this option should be set in server configuration

PR #625, opened 2015-03-12
My queue of various fixes

niol:
Details are in the commit messages.

### DIFF
--- a/helpers/ContentLoader.php
+++ b/helpers/ContentLoader.php
@@ -136,7 +136,7 @@ class ContentLoader {
             }
 
             // sanitize title
-            $title = $this->sanitizeContent($item->getTitle());
+            $title = $this->sanitizeField($item->getTitle());
             if(strlen(trim($title))==0)
                 $title = "[" . \F3::get('lang_no_title') . "]";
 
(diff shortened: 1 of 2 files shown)


Alcatr4zz:
Unfortunately, commit 1849b98 destroys links and line breaks in tweets:

Before:
before

After (not the same tweet):
after

niol:
OK, I understand that the twitter spout uses title for content. This is easy to fix : as I do not have a twitter API key, can you provide me with the elements to allow in the spout title? For your comment, I see ```a``` and ```br```. Are there any others?

Alcatr4zz:
No, there aren't any others, the `a` and `br` elements should be sufficient.

Alcatr4zz:
Thank you. Could you please add the `target` attribute to the allowed ones? This allows links to be opened in a new tab/window.

jtojnar:
Hmm, f49e1ae11b3aa2dac50ebdf32e2dd278b8941cc8 was essentially reverted by 7bb268925698e724432be7c82eeaa384e7856297, while still allowing the width attributes on random tags like tables. Any idea why would @SSilence want to revert it?

PR #627, opened 2015-03-15
Remove runtime asset building in favour of grunt

jtojnar:
Grunt is already used for packaging so why not let it build assets too. Less complex app is better.

Suggestions are welcomed.


gerundt:
Mhh, this would be a large barrier for *normal* user. :(

Modern web-developer should have no problem, but I think many users just have a FTP program and and a text editor.

jtojnar:
That’s what the package (http://selfoss.aditu.de/selfoss-2.12.zip) is for. Unless they change the assets or use development version they should make do without grunt.

Alternatively some build service could be set up (e.g. using travis).

gerundt:
Sorry for the misunderstanding. So I think this could be a good change! :+1:

niol:
Grunt is (still) not available in apt (Debian/Ubuntu) and that's what I use, so it would make it a little bit harder for my particular case to contribute. But I'll manage. This was for a very minor comment.

What is missing from this pull request though is some documentation and error messages for people who want to run the development version:
- uploading to a PHP enabled dir will silently fail. Maybe selfoss should display an error message in this case, testing if CSS and JS have been built.
- the installation section of the doc should be updated in case one wants to run the development version
- the update section should be updated in any case because step 5 is wrong if grunt builds the files.

jtojnar:
I will update the docs + add the error message later.

When on Ubuntu I installed it through npm from [Chris Lea’s PPA](https://launchpad.net/~chris-lea/+archive/ubuntu/node.js), not sure how compatible it is with Debian.

alfredbuehler:
I prefer the current state, don't see any real advantage in changing this.

SSilence:
Generally I would agree is the usage of grunt a good option. But in this case it would make the development of selfoss more difficult. Currently you can set DEBUG=1 in the config and the caching of js and css will be disabled. Debugging and programming new features is easier then with this extra step and extra toolchain (npm, grunt, ...). I use grunt currently only for generating the zip file. Other developers have not use grunt.

But thanks a lot for this pr!!

Issue #581, opened 2014-10-08
Unvalid auth


gllmhyt:
As said under commit 9d3d04bdf5f0dba157166d71f61d6dd9fa930333, there's some problems with this patch on my Nginx / PHP-FPM combo. Over HTTPS, it's working fine : I can login, and so on, with no problem. But over HTTP, having deleted the cookie related and purged the cache, I cannot login.

``` log
10-08-14 12:55:36 Debug session does not contain valid auth
```

elamperti:
I'm still having problems with this, using a similar server setup. `$_SERVER['HTTPS']` is set but empty.

PR #634, opened 2015-04-03
Fix login auth loop due to cookie not being set over HTTP

elamperti:
Under certain conditions the server may set `$_SERVER['HTTPS']` but it will be empty ([this is expected](http://php.net/manual/en/reserved.variables.server.php)) and it would be `"off"` only in some cases.

Because of this, the cookie isn't set and the user experiences a login loop with no error whatsoever, as previously commented on issue #581 and [on the selfoss support forum](http://selfoss.aditu.de/forum/index.php?id=530).

This pull request fixes the issue checking if `$_SERVER['HTTPS']` is empty.

### DIFF
--- a/helpers/Authentication.php
+++ b/helpers/Authentication.php
@@ -28,7 +28,7 @@ class Authentication {
     
         // session cookie will be valid for one month.
         $cookie_expire = 3600*24*30;
-        $cookie_secure = isset($_SERVER['HTTPS']) && "off"!==$_SERVER['HTTPS'];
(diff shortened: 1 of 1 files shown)


jtojnar:
Hi @elamperti. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions here and in other PRs under `GPL-3-or-later`?

elamperti:
Hi @jtojnar, I'm ok with licensing all my contributions (past and future) to fossar/selfoss as GPL3 or the license the author(s) find adequate :+1:

PR #644, opened 2015-04-27
allow target attribute

Alcatr4zz:
Allowing the target attribute within sanitizeField enables hyperlinks in titles to be opened in a new browser tab. This is especially useful for links within tweets.

### DIFF
--- a/helpers/ContentLoader.php
+++ b/helpers/ContentLoader.php
@@ -251,7 +251,7 @@ class ContentLoader {
         return htmLawed(
             htmlspecialchars_decode($value),
             array(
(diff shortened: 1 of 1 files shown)


bendem:
Wouldn't it better to set them all to _blank or removed them based on config?

Alcatr4zz:
I don't see why having this option would be useful. When clinking on a link within selfoss I expect it to be opened in a new tab instead of the selfoss tab and thus destroying my selfoss session.
As for the twitter spout, the links are already set to target=_blank but that's quite useless when the target attribute isn't allowed.

PR #647, opened 2015-05-06
Update .gitignore

kolydart:
Add custom file with javascript: user.js


kolydart:
Add custom file with javascript: user.js

PR #649, opened 2015-05-09
Config from env

cimnine:
I am preparing to run selfoss in a PaaS like Heroku. For that it must be possible to configure the app - at least partially - through ENV variables.

Even if I would bundle it up using Docker with configuration files, one would have to build a new Docker image upon mine to have his/her own config. (E.g. to change the salt, the username/password, the db credentials and so on)

That's why I implemented an easy way to set selfoss config variables though ENV variables. To prevent accidental conflicts the variables, by default, must be prefixed with a configurable prefix. I have pre-defined that prefix in defaults.ini to be "SELFOSS_". (The prefix check is case insensitive and the variable will be converted to lower case before adding it to f3.)

It would be nice to see this change integrated into the main stream.

### DIFF
--- a/common.php
+++ b/common.php
@@ -16,6 +16,14 @@ $f3->config('defaults.ini');
 if(file_exists('config.ini'))
     $f3->config('config.ini');
 
+// overwrite config with ENV variables
+foreach($f3->get('ENV') as $key => $value) {
+    $env_prefix = $f3->get('env_prefix');
+    if(strncasecmp($key,$env_prefix,strlen($env_prefix)) == 0) {
+        $f3->set(strtolower(substr($key,strlen($env_prefix))),$value);
+    }
+}
+
 // init logger
 $f3->set(
     'logger',
--- a/common.php
+++ b/common.php
@@ -17,8 +17,8 @@ if(file_exists('config.ini'))
     $f3->config('config.ini');
 
 // overwrite config with ENV variables
+$env_prefix = $f3->get('env_prefix');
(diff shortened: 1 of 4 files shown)


common.php
```diff
@@ -16,6 +16,14 @@
 if(file_exists('config.ini'))
     $f3->config('config.ini');
 
+// overwrite config with ENV variables
+foreach($f3->get('ENV') as $key => $value) {
+    $env_prefix = $f3->get('env_prefix');
```
bendem: That might be better to put this before the loop

cimnine: Good catch!

SSilence:
Thanks a lot for this pr. Did selfoss run on your heroku instance? A short howto in the wiki https://github.com/SSilence/selfoss/wiki would be great!

Thanks again
 Tobi

cimnine:
It run's on my on-premis PaaS built with [deis](http://deis.io), which is Heroku-like.

Most probably I will do another PR with my Procfile, once I figured out how to trigger the update in an elegant way.

cimnine:
See https://github.com/SSilence/selfoss/wiki/Run-selfoss-at-a-Heroku-like-PaaS. Cheers, Chris

jtojnar:
Hi @cimnine. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions here and in other PRs under `GPL-3-or-later`?

cimnine:
Yes, fine.

Issue #22, opened 2012-07-18
Internal Server Error / Undefined index: magic_quotes_gpc 


jbd:
Hello,

I've just installed selfoss on an apache server which is also running several php app without problem, and I've got an Internal Server Error : 

```
Undefined index: magic_quotes_gpc
#0 /selfoss/index.php:3 require('/selfoss/libs/f3/base.php')
```

Do you need to have magic_quotes_gpc turned on to run selfoss ?

PR #652, opened 2015-05-10
F3 would not launch when "magic_quotes_gpc" is not set.

cimnine:
Although magic_quotes_gpc was removed from recent PHP versions, f3 would not start without this property set. This patch fixes setups where this property is not set.

And because it is deprecated or removed, it does not matter if we overwrite it accidentally.

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -21,6 +21,11 @@ RewriteRule !public - [C]
 RewriteRule !index.php - [C]
 RewriteRule (.*) public/$1 [L]
 
+# set magic_quotes_gpc as f3 wouldn't start otherwise
+<IfModule mod_php5.c>
+    php_value magic_quotes_gpc "1"
+</IfModule>
+
 # deny requests for config files
 <FilesMatch ".(ini)$">
     Order allow,deny


jtojnar:
Are you sure #22 is occurring? I don't have `magic_quotes_gpc` set and selfoss works just fine.

cimnine:
Yes, I had exactly #22 today. PHP 5.6.7 on Apache 2.4.10 w/ selfoss/master@HEAD (i.e. 739e4a29b130dc6a8ff3c774090a7d0506c1d36b).

jtojnar:
I checked it out but it still works for me.
```
➜  selfoss git:(739e4a2) ✗ php -v
PHP 5.6.8 (cli) (built: Apr 20 2015 09:30:53) 
Copyright (c) 1997-2015 The PHP Group
Zend Engine v2.6.0, Copyright (c) 1998-2015 Zend Technologies
➜  selfoss git:(739e4a2) ✗ apachectl -v
Server version: Apache/2.4.12 (Unix)
Server built:   Jan 31 2015 11:02:34
➜  selfoss git:(739e4a2) ✗ php -r "var_dump(ini_get('magic_quotes_gpc'));"
bool(false)
```

jtojnar:
That’s just weird. Wouldn’t, however, be better to patch f3 directly?

cimnine:
I removed it from my config and it does not happen now. That's very weird. Probably you are right and this should be fixed in f3.

jtojnar:
Anyway, it should be set to "0", if ever. Having `magic_quotes_gpc` enabled is never a good thing. Even fatfree tries to set it to 0 too (https://github.com/SSilence/selfoss/blob/master/libs/f3/base.php#L1606)

PR #657, opened 2015-05-13
Add link to detailed example installation instructions

Poorchop:


jtojnar:
Hi @Poorchop. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions here and in other PRs under `GPL-3-or-later`?

Poorchop:
Yes, that is fine with me. Thanks.

PR #659, opened 2015-05-13
Bump version in pgsql/Database.php

### LINKED ISSUE #656: Error PDOStatement: ERROR:  column "filter" of relation "sources" already exists
Using PostgreSQL as a backend, Selfoss 2.13 runs once but always crashes after that.  
Workaround:
* Commenting out lines 178-183 in daos/pgsql/Database.php
* Manually setting the version row to 6

jeffcasavant:
Fixes #656

### DIFF
--- a/daos/pgsql/Database.php
+++ b/daos/pgsql/Database.php
@@ -114,7 +114,7 @@ class Database {
                 ');
                 
                 \F3::get('db')->exec('
-                    INSERT INTO version (version) VALUES (5);
+                    INSERT INTO version (version) VALUES (6);
                 ');
                 
                 \F3::get('db')->exec('


jtojnar:
This should work. I couldn’t decide whether to change the version here or rather save it to a constant and place it in [daos/Database.php](https://github.com/SSilence/selfoss/blob/2.13/daos/Database.php). That would make it more difficult to forget to change it.

SQLITE database is, however, one version ahead because of compatibility problems (b74a13556e5b0c3d55120f7fc258d4bb50b3fcb4) and I am not sure what to do about it. Probably it won’t matter if we skip a version to sync it.

jeffcasavant:
https://github.com/SSilence/selfoss/issues/656#issuecomment-101741638 Niol brought up a good point - in the case in which someone is upgrading from 5, does this fix the whole issue? I've kept up the discussion over there.

jtojnar:
Yes. This issue only occurs with fresh installs, because they are thought to be version 5 when they actually aren’t. Version 5 lacks the column so it can be upgraded safely.

jeffcasavant:
Who can merge this?

jtojnar:
@SSilence only, I think.

Issue #643, opened 2015-04-27
Random internal server errors after adding some feeds


Poorchop:
I tried really hard to get selfoss working on my server and after a lot of messing around, things finally seemed to be up and running. However, it seems that after adding certain feeds, things go haywire and selfoss breaks down completely. I added a GitHub feed and some comic feeds, but after adding an NPR feed, I kept getting the following error when trying to manually refresh (IP and server information redacted):

`[error] 2155#0: *57 FastCGI sent in stderr: "PHP message: Fatal error: Call to undefined function imagecreatefromstring()" while reading response header from upstream, client: [IP], server: [server], request: "GET /update HTTP/1.1", upstream: "fastcgi://127.0.0.1:9000", host: "[server]"`

Now when I try to delete the NPR feed or any feed for that matter, I get a message in my web browser that says "Error deleting source: Internal Server Error" and the following error in my logs:

`[error] 2155#0: *57 FastCGI sent in stderr: "PHP message: Fatal error: Automatically populating $HTTP_RAW_POST_DATA is deprecated and will be removed in a future version. To avoid this warning set 'always_populate_raw_post_data' to '-1' in php.ini and use the php://input stream instead" while reading response header from upstream, client: [IP], server: [server], request: "POST /source/delete/3 HTTP/1.1", upstream: "fastcgi://127.0.0.1:9000", host: "[server]"`

I'm guessing the second error message is kind of irrelevant to the issue at hand.

I've also noticed that even when selfoss works initially before everything breaks down, it doesn't display any favicons or thumbnails. Data/favicons and data/thumbnails are always empty after adding feeds, so clearly some things are amiss. I really wish I knew what I was going wrong because I just can't seem to get selfoss completely working.

I'm using nginx and sqlite on Arch.

niol:
For the ```imagecreatefromstring``` , can you help me reproduce by providing spout info?

For the cannot edit source problem, this is the same as #565 .

Poorchop:
I am using the default "RSS Feeds" spout. I can successfully add my personal GitHub feed and [this feed](https://feeds.feedburner.com/PoorlyDrawnLines). Once I added the [NPR News feed](http://www.npr.org/rss/rss.php?id=1001) and I tried to refresh, that's when I started getting error messages and had to reinstall from scratch.

[...]

Issue #565, opened 2014-09-25
Cannot Star/Unstar : Internal Error


Keyhaku:
Hi,

I've just met a problem with the star/unstar system, see my nginx log, by uncommenting the line
";always_populate_raw_post_data = -1" in php5.ini i've managed to make it work.

According to php manual (http://php.net/manual/en/ini.core.php), this directive is obsolete from php 5.6 (that the version I'm using).

Regards.
## Log

*182 FastCGI sent in stderr: "PHP message: Fatal error: Automatically populating $HTTP_RAW_POST_DATA is deprecated and will be removed in a future version. To avoid this warning set 'always_populate_raw_post_data' to '-1' in php.ini and use the php://input stream instead"

jtojnar:
As temporary fix you can replace line 1757 in [libs/f3/base.php](https://github.com/SSilence/selfoss/blob/master/libs/f3/base.php#L1757)

		if ($error=error_get_last())

with

		$ignored = "Automatically populating \$HTTP_RAW_POST_DATA is deprecated and will be removed in a future version. To avoid this warning set 'always_populate_raw_post_data' to '-1' in php.ini and use the php://input stream instead.";
		if ($error=error_get_last() && isset($error) && $error['message']!==$ignored)

The issue is described in detail in this  [blog post](https://www.bram.us/2014/10/26/php-5-6-automatically-populating-http_raw_post_data-is-deprecated-and-will-be-removed-in-a-future-version/).

Maybe we should also start an issue in https://github.com/bcosca/fatfree

jtojnar:
Closing since it was fixed in #664. Please open a new issue when it reappears.

PR #664, opened 2015-05-19
"$HTTP_RAW_POST_DATA is deprecated" error

fmunch:
As of PHP 5.6, `$HTTP_RAW_POST_DATA` is deprecated. Sending a POST request without Content-Type populates `$HTTP_RAW_POST_DATA`, which triggers an `E_DEPRECATED`.

It makes the Star / Unstar button fail:

Request:
```
POST /selfoss/starr/26110 HTTP/1.1
Host: ...
Connection: keep-alive
Content-Length: 0
Accept: */*
Origin: ...
X-Requested-With: XMLHttpRequest
User-Agent: Mozilla/5.0 (Windows NT 6.3; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/44.0.2398.0 Safari/537.36
DNT: 1
Referer: ...
Accept-Encoding: gzip, deflate
Accept-Language: fr,en-US;q=0.8,en;q=0.6
Cookie: PHPSESSID=...
```

Response headers:
```
HTTP/1.1 500 Internal Server Error
Server: nginx/1.6.0
Date: Tue, 19 May 2015 07:40:49 GMT
Content-Type: text/html; charset=UTF-8
Transfer-Encoding: chunked
Connection: keep-alive
X-Powered-By: PHP/5.6.8

{"status":"Internal Server Error","code":500,"text":"Fatal error: Automatically populating $HTTP_RAW_POST_DATA is deprecated and will be removed in a future version. To avoid this warning set 'always_populate_raw_post_data' to '-1' in php.ini and use the php:\/\/input stream instead.","trace":[]}
```

A `data` key was added with `{ ajax: true }`, in order to be consistent with the second AJAX call in selfoss-events-entriestoolbar.js (line 182). Adding `contentType: 'application/x-www-form-urlencoded; charset=UTF-8'` would work too.

### DIFF
--- a/public/js/selfoss-events-entriestoolbar.js
+++ b/public/js/selfoss-events-entriestoolbar.js
@@ -90,6 +90,7 @@ selfoss.events.entriesToolbar = function(parent) {
             
             $.ajax({
                 url: $('base').attr('href') + (starr ? 'starr/' : 'unstarr/') + id,
+                data: { ajax: true },
                 type: 'POST',
                 error: function(jqXHR, textStatus, errorThrown) {
                     // rollback ui changes


niol:
Does this fix #565?

fmunch:
You're right, I wasn't aware of #565 when opening the pull request.
It should fix it.

fmunch:
However it does not fix #643, obviously.

niol:
It should solve the second part of #643. The first part is fixed.

Issue #587, opened 2014-10-13
Refresh button AJAX call gets a 404


niol:
The AJAX call introduced by #394 gives me a 404. Isn't the patch missing a route in ```index.php``` or something in ```.htaccess``` ? Or is something wrong with my PHP configuration?

I could not find any documentation addressing this.

PR #670, opened 2015-05-30
Changed the update url for the refresh button

aminecmi:
It needs to be changed in the "all-v2.13.js" too.

### DIFF
--- a/public/js/selfoss-events-navigation.js
+++ b/public/js/selfoss-events-navigation.js
@@ -149,7 +149,7 @@ selfoss.events.navigation = function() {
         $('#content').addClass('loading').html("");
           
         $.ajax({
-            url: $('base').attr('href') + 'update',
+            url: $('base').attr('href') + 'Update',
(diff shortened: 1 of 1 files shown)


bendem:
[Update route](https://github.com/SSilence/selfoss/blob/master/index.php#L59) is lower case, what's the reason behind this change?

aminecmi:
The lower case update route returns a 404 error.
I seen on the forum people saying that it was with an upper case U, and it seems to work with theese changes.

niol:
Maybe this is the same as #587 .

aminecmi:
Disabling the MultiViews option didn't elites

niol:
Disabling the MultiViews option solved the same problem for me. What happens if your rename ```update.php```?

aminecmi:
It turns out, Multiviews was enabled for a subdirectory. Changing this resolved the problem.
Can you add theses informations to the main website ?

PR #674, opened 2015-06-05
fix blockquote and pre overflow-x resetting

mcnesium:
blockquote and pre's used to be stuck on one line, this fixed it


jtojnar:
Do you recall what you meant by stuck on one line? Also the `pre` part was reverted in #889.

PR #679, opened 2015-06-27
Add functionality to show when the last post was found for a source.

martinsauter:
In the "(add) sources" list in the web browser the number of days since a new post was 
found for each source will be shown to the user. 

The functionality is implemented by adding a new field to the sources table into which 
the date/time is written when a new post was found for the source.


aslmx:
Nice idea as it will help clean up the sources..

but...

i would need some googling for this, but is the extra column required? I would have thought the datetime for the last posting of each source could be simply fetched via some tricky sql. 

vgs

niol:
Yes it would be simple to join with the items table. But the proposed implementation here uses the date of the latest insertion, which is different from the date of the latest post.

martinsauter:
I thought about using the information contained in the items table at first, too. Unfortunately this approach does not work because if you don't "star" a post of a source there will be no posts of a source after the post expiry time (default 30 days).. 

I have a lot of sources with more than 30 days in between but I would still like to know the time the last post was seen even if there is currently no post of that source in the database (e.g. to delete the source if I haven't seen anything for 6 months). The only way to do this is to have an extra field in the sources table that stores the last time a post was found as this information does not get delete when the last post of that source is removed from the items table.

aslmx:
right, i forgot about the automatic cleaning which is quite likely to happen for a source that does not publish very often.

Then the new column seems to be the only option.

helpers/ContentLoader.php
```diff
@@ -185,6 +187,8 @@ public function fetch($source) {
             
             \F3::get('logger')->log('Memory usage: '.memory_get_usage(), \DEBUG);
             \F3::get('logger')->log('Memory peak usage: '.memory_get_peak_usage(), \DEBUG);
+            
+            $newPostAdded = TRUE;
```
PhrozenByte: Instead of an boolean ```$newPostAdded``` parameter you could use something similar to ```$lastEntryDate = max($lastEntryDate, $itemDate->getTimestamp());```. Lines 61-64 would then be ```$lastEntryDate = $source['lastentry']; $this->updateSource($source, null);```.

SSilence:
Thanks a lot for this pr. A really good idea. But it doesn't work correctly. On following feed it shows 25 days, but last post is 6 hours ago:
http://t3n.de/tag/javascript.atom

Any idea?

Thanks a lot
 Tobi

daos/sqlite/Database.php
```diff
@@ -197,4 +211,27 @@ public function optimize() {
...
+     * @return void
+     */
+    private function initLastEntryFieldDuringUpgrade() { 	
+    	$sources = @\F3::get('db')->exec('SELECT id FROM sources');
+   
+    	// have a look at each entry in the source table
+    	foreach($sources as $current_src) {
+    		//get the date of the newest entry found in the database   		
+    		$latestEntryDate = @\F3::get('db')->exec('SELECT datetime FROM items WHERE source=' . 
+    				                              $current_src['id'] . ' ORDER BY id DESC LIMIT 0, 1');
```
PhrozenByte: Should be ```ORDER BY datetime```

martinsauter: Hi, yes I can confirm, ordering by datetime fixes the problem! 

The only scenario I can imagine at the moment when the ordering by id would fail is when the RSS feed was added AFTER the latest post was published and BEFORE running the database upgrade. To stick with your example you must have added the feed today after 13:05 and then implicitly run the database upgrade. 

Is this what happened (or am I missing another scenario?).

If you like I can do a pull request with the fix or you can insert it yourself, as you like.

Thanks to both of you for spotting this!
Martin

SSilence:
Ah yes, I see thats the reason. After fetching one new item the issue will be fixed automatically.

Thanks a lot!
 Tobi

martinsauter:
O.k. cool! Still worth to be fixed. I just played around with my database, deleted the new field, went back to v6 and ran the original and suggested replacement queries. Replacing "BY id" with "BY datetime" fixes the issue nicely.

PR #684, opened 2015-07-06
Refactor shares to only be in javascript so it is defined once and mo…

### LINKED ISSUE #680: Generic Share buttons via config.ini
This is a feature request - but also a thread discussing how to implement it in case i will manage to do it eventually myself in case no one else does ;-)

I just implemented myself a "Download PDF" button which simply sends the current item source to a webservice that converts the page into a nice pdf (for your reference: www.printfriendly.com, but there are plenty of others). Works quite well, but the effort is quite high for the average user.

I thought about adding the addition to my branch and pull request it, but i first thought i'd discuss whehter it is possible to find a generic solution which lets the user configure ANY share button via the config.ini.

I had to change selfoss at 4 places to make the extra button. I'm not that experienced with state of the art javascript/web development and did not yet fully understand the way how selfoss sourcecode actually works.

How would you best implement this generic share buttons? I thought the feature should look like this in the config.ini
example: share to pdf

sharebutton_name1=Share to PDF
sharebutton_key1=p
sharebutton_image1=images/buttonpdf.png
sharebutton_url1=http://thefanciestpdfserviceintheworld.com/?url=

so whatever piece of code could simpyl try to iterate of the sharebutton_***x keys in the config.ini and "render" the buttons accordingly.

My problems are: how to start the iteration?
e.g. the Image/icon for the buttons are declared in a .css file which is not processed by PHP
Or the javacscript which actually puts the functionality to the buttons. How could i iterate the configuration here?

I would really appreciate if someone with more experience would simply implement this, but i'm open for sugesstions and then try to do it myself... 

I think there are numerous use cases. e.g. there is also a webservice that converts a webpage to epub and then sends it do your Amazon Kindle (i never tried it because i'm converting them to PDF and then send them to my kindle myself ;-) and maybe other things like rarely used social networks that let users share URLs... 

thanks for your help!
br
aslmx

bbeardsley:
…re easily  (removed specific share code from css/php).  Can add new shares more simply now by adding code to user.js and adding an png image. This fixes #680 (a little differently than requested - in javascript instead of config.ini).


bbeardsley:
Hmmm, found an issue when opening and closing the same item it adds duplicate share buttons...

bbeardsley:
Fixed the conflicts.

SSilence:
thanks for this pr and fixing the merge conflicts!

bbeardsley:
No problem.  Was wondering where I should document (other than #680) how you can easily add your own custom shares by modifying user.js and adding an image?

jtojnar:
Hi @bbeardsley. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions here and in other PRs under `GPL-3-or-later`?

bbeardsley:
I'm fine with that.

PR #707, opened 2015-09-27
better handling of load list errors

Kleofass:


jtojnar:
Hi @Kleofass. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions here and in other PRs under `GPL-3-or-later`?

Kleofass:
No problem, I accept.

PR #709, opened 2015-09-30
Add spout "RSS Feed (with enclosures)" to support feed enclosures

PhrozenByte:
According to the SimplePie docs, the `get_enclosures()` method...
> Supports the \<enclosure> RSS tag, as well as Media RSS and iTunes RSS.

Example: https://erzaehlmirnix.wordpress.com/feed/

### DIFF
--- a/spouts/rss/enclosures.php
+++ b/spouts/rss/enclosures.php
@@ -1,0 +1,51 @@
+<?PHP 
+
+namespace spouts\rss;
+
+/**
+ * Plugin for fetching RSS feeds with image enclosures
+ *
+ * @package    plugins
+ * @subpackage news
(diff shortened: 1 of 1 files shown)


aslmx:
Funny... i was just about to raise the question whether and when media will be supported...
Even more funny, that i was gonig to post the exact same example Rss feed ;-)

nice.
Thank you very much for your sprout.... works like a charm! :+1:

jtojnar:
Hi @PhrozenByte. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions here and in other PRs under `GPL-3-or-later`?

PR #712, opened 2015-10-20
Add keyboard shortcut to refresh sources (Shift+r)

Sudiukil:
Can't help to think there is probably a good reason for the lack of such a shortcut.

I added it anyway, can be useful...

Any thoughts on this ?


jtojnar:
Hi @Sudiukil. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions under `GPL-3-or-later`?

PR #714, opened 2015-10-22
Add support for private (@) and hidden (#) tags

wazari972:
Hello,

I had these patches in my version of selfoss since quite a while, and I wanted to suggest it to you.
It add the ability 

1/ to have private tags (with a @ in the tag name). Items (and sources) with such tags won't be visible in the feed if the user is not logged in.

2/ to have "not important" tags (with a # in the tag name), whose items won't be visible in the "all tags" feed, but only when you click directly on the tag or source.

Currently the # and @ symbols are hardcoded, but I can update the code to allow the configuration in defaults.ini (or even disabling it with an empty value).

### DIFF
--- a/controllers/Index.php
+++ b/controllers/Index.php
@@ -49,7 +49,16 @@ class Index extends BaseController {
         $this->view->statsAll = $stats['total'];
         $this->view->statsUnread = $stats['unread'];
         $this->view->statsStarred = $stats['starred'];
-        
+
+        if ($tagsDao->hasTag("#")) {
+		foreach ($tags as $tag) {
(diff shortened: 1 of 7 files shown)


jtojnar:
Hi @wazari972. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contribution under `GPL-3-or-later`?

kpouget:
Hello @jtojnar yes I agree with licensing my contribution under `GPL-3-or-later`

thanks for the work you're going,

Kevin

PR #716, opened 2015-10-27
Remove joins from daos/mysql/Items.php

jbfavre:
When you want to scale an application, it's sometimes better to not use JOIN statements.
Many fast SQL queries are often better than a slow one.

That's exactly the case here where, on a small server, JOIN operations will systematically use tmp tables. In that case, it's better to do the join in PHP.

More, there are absolutely no need for a JOIN just to check whether we have more items

### DIFF
--- a/daos/mysql/Items.php
+++ b/daos/mysql/Items.php
@@ -265,10 +265,6 @@ class Items extends Database {
             $source_list = array();
             foreach ($resultset as $source)
                 $sources_list[] = $source['id'];
-            ob_start();
-            var_dump(array_values($sources_list));
-            $result = ob_get_clean();
-            \F3::get('logger')->log('sources_list for tag'.$options['tag'].': '.$result, \ERROR);
             $where .= " AND source in (".implode(',',$sources_list).")";
         }
 
--- a/daos/mysql/Items.php
+++ b/daos/mysql/Items.php
@@ -78,8 +78,44 @@ class Items extends Database {
         \F3::get('db')->exec('UPDATE '.\F3::get('db_prefix').'items SET '.$this->stmt->isFalse('starred').' WHERE id=:id',
                     array(':id' => $id));
     }
-    
-    
+
+
+    /**
+     * mark item as shared (used for source scoring)
+     *
+     * @return void
+     * @param int $id the item
+     */
+    public function shared($id) {
+        \F3::get('db')->exec('UPDATE '.\F3::get('db_prefix').'items SET '.$this->stmt->isTrue('shared').' WHERE id=:id',
+                    array(':id' => $id));
+    }
+
+
+    /**
+     * unshare item (used for source scoring)
+     *
+     * @return void
+     * @param int $id the item
+     */
+    public function unshared($id) {
+        \F3::get('db')->exec('UPDATE '.\F3::get('db_prefix').'items SET '.$this->stmt->isFalse('shared').' WHERE id=:id',
+                    array(':id' => $id));
+    }
+
+
+    /**
+     * mark item as opened (used for source scoring)
+     *
+     * @return void
+     * @param int $id the item
+     */
+    public function opened($id) {
+        \F3::get('db')->exec('UPDATE '.\F3::get('db_prefix').'items SET opened=1 WHERE id=:id',
+                    array(':id' => $id));
+    }
+
+
     /**
      * add new item
      *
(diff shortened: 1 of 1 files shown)


niol:
Do you have benchmarks supporting this?
Did you test with sqlite and pgsql?
Why count in PHP rather than in SQL?

jbfavre:
The only feedback I have is my own instance with >500 sources & >250000 items.
Got many timeouts while loading items (30 seconds timeout). Almost none since I modify the code.

Sorry, I forgot to update sqlite & pgsql part. Will do it and update PR
I haven't tested with those 2 DBs since I only use MySQL with full dataset (and besides, I'm pretty sure it's not safe to load 250000 items into sqlite).

I do not count in PHP since the original request did not count.
I only perform the equivalent of a join in PHP. That makes more requests, but each one is faster and PHP code is also quite fast.

This also allows to decrease timeouts when updating web UI during a feed update, and it also improve concurrent access because lock time wait is reduced

jbfavre:
Just figured out sqlite & pgsql Items daos uses mysql one.
No need to update though.

I can only say my code works with sqlite. Have no idea about performance since I only use sqlite for dev tests with few sources & items only.

Have no way to test with pgsql since I don't use it, nor know it.

niol:
Your patch breaks pgsql (```LIMIT 15,1``` should be ```LIMIT 1 OFFSET 15```). If I fix this, I cannot see any difference with my 130k items and 40 sources.

```
Tue, 27 Oct 2015 22:14:29 +0000 (1,3ms) SELECT items.id
                   FROM items AS items, sources AS sources
                   WHERE items.source=sources.id  AND unread=true 
                   LIMIT 1 OFFSET 15
Tue, 27 Oct 2015 22:14:29 +0000 (109,6ms) SELECT 
                    items.id, datetime, items.title AS title, content, unread, starred, source, thumbnail, icon, uid, link, updatetime, author, sources.title as sourcetitle, sources.tags as tags
                   FROM items AS items, sources AS sources
                   WHERE items.source=sources.id  AND unread=true 
                   ORDER BY items.datetime ASC 
                   LIMIT 15 OFFSET 0
```

```
Tue, 27 Oct 2015 22:13:18 +0000 (0,9ms) SELECT items.id
                   FROM items
                   WHERE 1=1  AND unread=true
                   LIMIT 1 OFFSET 15
Tue, 27 Oct 2015 22:13:18 +0000 (109,0ms) SELECT 
                    id, datetime, title, content, unread, starred, source, thumbnail, icon, uid, link, updatetime, author
                   FROM items AS items
                   WHERE 1=1  AND unread=true 
                   ORDER BY datetime ASC 
                   LIMIT 15 OFFSET 0
Tue, 27 Oct 2015 22:13:18 +0000 (0,8ms) SELECT 
                        title as sourcetitle, tags
                       FROM sources
                       WHERE id=36
Tue, 27 Oct 2015 22:13:18 +0000 (0,7ms) SELECT 
                        title as sourcetitle, tags
                       FROM sources
                       WHERE id=8
Tue, 27 Oct 2015 22:13:18 +0000 (0,6ms) SELECT 
                        title as sourcetitle, tags
                       FROM sources
                       WHERE id=26
Tue, 27 Oct 2015 22:13:18 +0000 (0,6ms) SELECT 
                        title as sourcetitle, tags
                       FROM sources
                       WHERE id=29
Tue, 27 Oct 2015 22:13:18 +0000 (0,6ms) SELECT 
                        title as sourcetitle, tags
                       FROM sources
                       WHERE id=23
Tue, 27 Oct 2015 22:13:18 +0000 (0,6ms) SELECT 
                        title as sourcetitle, tags
                       FROM sources
                       WHERE id=97
```

Can you show some logs on how this is helping mysql? The above log was obtained by appending ```echo \F3::get('db')->log();``` at the end of ```index.php```.

Also, your patch breaks using mysql with a table prefix, you are missing a ```AS items``` in the ```FROM``` clause of your first fixed query.

I do not have a big mysql db to test... and it'll be a few days before I have one, so I won't be able to test soon.

jbfavre:
I just figured it also breaks tag filtering. I have to work a bit more on it.

Will provide you required logs

jbfavre:
For the record, I run selfoss on a little server with only 2GB of memory.
In this setup, MySQL only have 1GB available. That may explain why I see slowness & timeouts :)

niol:
The logs are not *required*, I just want to understand what you are trying to fix, If it is the right fix and if it is mysql specific.

jbfavre:
Yes, I choose the wrong word sorry.

daos/mysql/Items.php
```diff
@@ -246,18 +246,36 @@ public function get($options = array()) {
         
         // first check whether more items are available
         $result = \F3::get('db')->exec('SELECT items.id
-                   FROM '.\F3::get('db_prefix').'items AS items, '.\F3::get('db_prefix').'sources AS sources
-                   WHERE items.source=sources.id '.$where.' 
-                   LIMIT 1 OFFSET ' . ($options['offset']+$options['items']), $params);
+                   FROM '.\F3::get('db_prefix').'items
+                   WHERE 1=1 '.$where.'
+                   LIMIT ' . ($options['offset']+$options['items']) . ', 1', $params);
         $this->hasMore = count($result);
```
PhrozenByte: Using a `COUNT()` SQL Query isn't necessarily faster, but definitely requires much less memory.

daos/mysql/Items.php
```diff
@@ -246,18 +246,36 @@ public function get($options = array()) {
...
+        // from databse only if needed (ie. if not already retrieved)
+        $sources_list = array();
+        foreach ($items_list as $key => $item) {
+            if (array_key_exists($item['source'], $sources_list)) {
+                $items_list[$key] = array_merge($items_list[$key], $sources_list[$item['source']]);
+            }else{
+                $source = \F3::get('db')->exec('SELECT 
+                        title as sourcetitle, tags
+                       FROM '.\F3::get('db_prefix').'sources
+                       WHERE id='.$item['source']);
```
PhrozenByte: Collecting all `$item['source']` IDs first and then executing a single `WHERE id IN (...)` SQL query is much faster than multiple queries.

jbfavre:
@niol sorry for not updating sooner, here is the output of bd->log() command as asked:

niol:
And without your optimizations?

niol:
If I count correctly, the JOIN is not that much faster (mySQL time) than the multiple SELECTs : 2.3ms + 21 * 0.3 vs 10.6ms. With PHP time, that gap should be narrower.

What is strange is that although the statement is the same, the last query is very long (more than 2s).

I'll build myself a bug mysql database in the next few days to help me make sense of this.

jbfavre:
I think difference of time for the same statement can be quite easily explained: lock.
That's precisely the reason why I offer to make many tiny, and fast, requests instead of few big, and slower, ones

niol:
Can you elaborate because I do not understand why a 10ms read-only query query would have another one wait for two seconds. And PHP is a single thread. And I cannot reproduce. And nothing is locking tables un selfoss. What makes you think about locks?

PR #726, opened 2015-12-19
Auto collapse

agnibho:
Added a new feature auto_collapse.
It can be used by setting `auto_collapse=1` in the config.ini file. By default the value is set to 0 and the feature is deactivated.
When activated the open items will collapse automatically when another item is opened, thereby making the interface even more clutter free and easy to browse.


jtojnar:
Hi @agnibho. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions under `GPL-3-or-later`?

PR #727, opened 2015-12-28
Add base font color

jtojnar:
If the user uses dark theme, the text can be illegible. When changing a background color, the font color should be changed too.

Light font color


f3bruary:
there's a dark theme ?

jtojnar:
Operating system’s dark theme. In Firefox, you can choose to use theme’s color scheme as default colors.
screenshot from 2015-12-28 18-18-53

f3bruary:
oooh it's a browser feature...

Issue #515, opened 2014-04-19
Composer and Packagist


luxifer:
Hey guys, did you have a look at [composer](http://getcomposer.org/) and [packagist](https://packagist.org/) to manage dependencies?
Is there a place where I can find an exhaustive list of dependencies used by this project so I can make a PR for a PoC.

Issue #636, opened 2015-04-10
Undefined offset: 28


rvimont:
Since yesterday, feed upgrade fails (without any change in my selfoss version/configuration).

With DEBUG mode enabled, here is the end of `data/logs/default.log`:
~~~
04-10-15 8:33:09 Debug ---
04-10-15 8:33:09 Debug start fetching source "Notes d'un &eacute;conomiste (id: 1177) 
04-10-15 8:33:09 Debug spout successfully loaded: spouts\rss\feed
04-10-15 8:33:09 Debug fetch content
04-10-15 8:33:10 Error error loading feed content for MANGE TA MAIN!: This XML document is invalid, likely due to invalid characters. XML error: Mismatched tag at line 59, column 8
04-10-15 8:33:10 Error error loading feed content for Notes d'un &eacute;conomiste: This XML document is invalid, likely due to invalid characters. XML error: Not well-formed (invalid token) at line 1, column 1
04-10-15 8:33:10 Debug ---
04-10-15 8:33:10 Debug start fetching source "Jean Zin (id: 1275) 
04-10-15 8:33:10 Debug spout successfully loaded: spouts\rss\feed
04-10-15 8:33:10 Debug fetch content
04-10-15 8:33:10 Error error loading feed content for Notes d'un &eacute;conomiste: This XML document is invalid, likely due to invalid characters. XML error: Not well-formed (invalid token) at line 1, column 1
04-10-15 8:33:12 Error Undefined offset: 28

04-10-15 8:33:12 Debug minimum date: 2015-04-05 08:33:12
04-10-15 8:33:12 Debug start item fetching
04-10-15 8:33:12 Debug start insertion of new item "Philosophie politique et politique effective"
04-10-15 8:33:12 Debug item content sanitized
04-10-15 8:33:13 Error Undefined offset: 28
~~~

niol:
This is a bug in WideImage in the BMP loading code of the favicon.

The following change fixes the problem. WideImage seems dead so I won't submit the fix unless someone tells me otherwise.

Can you confirm the fix so I can submit a PR?

```diff
--- a/libs/WideImage/vendor/de77/BMP.php
+++ b/libs/WideImage/vendor/de77/BMP.php
@@ -131,7 +131,7 @@ class WideImage_vendor_de77_BMP

                $bps = $header['bits_per_pixel']; //bits per pixel
            $wid2 = ceil(($bps/8 * $header['width']) / 4) * 4;
[...]

PR #735, opened 2016-01-14
[WIP] Composer 

jtojnar:
This is still work in progress.

As recommended in #515 I decided to use composer to manage dependencies.

[floIcon](http://www.phpclasses.org/package/3906-PHP-Read-and-write-images-from-ICO-files.html) that was used previously apparently has restrictive license and isn’t available on packagist. jimIcon claims to be a drop-in replacement, it needs to be tested though. I opened [a pull request](https://github.com/jimparis/jimIcon/pull/1) adding composer.json.

We can also get rid of submodules once this [pull request](https://github.com/fivefilters/ftr-site-config/pull/76) gets accepted.

Last remaining libraries not on composer are two FullTextRss components. Where do they come from? It seems like FullTextRss is a [commercial product](http://fivefilters.org/content-only/). For now, I moved them to `vendor/others` directory.


j0k3r:
For Readability, here is the original from FiveFilters: https://bitbucket.org/fivefilters/php-readability/src (using `fivefilters/php-readability` in your composer).

Otherwise I've forked a fork of that lib which are used in [Wallabag](https://github.com/wallabag/wallabag) project and other one. It's here https://github.com/j0k3r/php-readability

Regarding the ContentExtractor, it's a core part for full-text-rss ([see](https://bitbucket.org/fivefilters/full-text-rss/src/daedf214febe8d944141bb92e2a4c06b8e4a6c12/libraries/content-extractor/?at=master)) you won't be able to get it using composer.

niol:
- Be careful of bugs which were fixed in the selfoss forked libs but not in the original libs. I can think of #636 but there may be more. The diff of the deleted libs vs. the ones pulled from composer should be verified to prevent regressions.
- Please make sure an archive with all deps gets built for releases to make it easy to install selfoss releases

jtojnar:
@j0k3r Bitbucket stopped supporting custom domain names so `fivefilters/php-readability` [can’t be cloned](https://bitbucket.org/fivefilters/php-readability/pull-requests/1) currently. I also like your version better as it has at least some tests.

We could use ContentExtractor from your [Graby](https://github.com/j0k3r/graby), although Graby itself is probably a bit larger than needed. Could the Extractor be extracted to separate composer package? Or maybe selfoss could use Graby instead of SimplePie/ContentExtractor/… combo (another PR perhaps).

@niol that’s why errors should always be fixed upstream :laughing: I will go through https://github.com/SSilence/selfoss/commits/master/libs and try to send pull requests to original libraries if it isn’t fixed in them already. It will have to be verified nevertheless.

I will update Gruntfile later.

j0k3r:
@jtojnar Could be hard to extract. Maybe you should directly use graby instead of ContentExtractor, so you won't need php-readability since it's included in graby.

SSilence:
Thank you for this pr and improvement. I will have a look on composer next time and try to introduce this package manager. I'm not sure this is a good idea for selfoss, because of a few fixes in libraries. floIcon is a little class I copied and I think this should be handled like selfoss source code.

jtojnar:
@SSilence the fixes are already in upstream so that should not be a problem. The reason I did not finish this yet is that I could not find a reasonable way to include the readability filters.

Spy-Seth:
They could be added in a dedicated repository and added as dependence from there. It do not soldve completely the issue but allow to keep moving on this subject.

PR #738, opened 2016-01-19
Modify style.css

ZMOM1031:
Replace font styles, Add code syntax style


jtojnar:
Adding a styles for `<code>` is a good idea. I don’t understand the other changes, though.

I guess the new fonts are for better legibility with Chinese text, although I can’t compare that, as I don’t have *any* of the fonts you specified installed. Chinese falls back to Droid Sans (my system font) with both your fonts and selfoss’s bundled Open Sans.

What is the purpose of the darkened background? It lowers the contrast between text and and the background and may make reading the text little harder for some users.

Removing columns also makes the articles harder to read. Compare the following images:

latin text, new style
chinese text, new style
chinese text, old style

ZMOM1031:
Code Style Blog problem is my problem.
Use a dark background because white background look too bright at night, the eyes can not stand.
Remove Columns is because more content to watch from the beginning of the second column, look at the third column
I do not know why your content is not centered, I was under the 2.14 version of the reform
https://github.com/ZMOM1031/selfoss/releases/tag/2.14-1
selfoss_1
selfoss_2
selfoss_3
selfoss_4

jtojnar:
I prefer dark backgrounds too but I don’t think dark background is good for all users. You can set your personal CSS in `user.css` file. Maybe in future selfoss can have dark theme and light theme option.

user.css location

Your code has no image centring. I don’t understand why are your images centred. I tested your changes on 9e6347b.

The columns are for large screens. Removing the columns is not solution. Long lines are hard to read.

Some solutions:
* Something like [CSS figures](https://figures.spec.whatwg.org/)
* If article contains code, add class `entry-content-nocolumns`
* Add `overflow: auto` to code
   screenshot from 2016-01-19 19-43-40
   ```css
     .entry-content pre {
         overflow: auto;
     }

     .entry-content table { /* your blog wraps code in table */
         display: block;
         overflow: auto;
     }
    ```

Long lines of text are hard to read:
screenshot from 2016-01-19 18-53-50
Images are not centred:
screenshot from 2016-01-19 19-11-42

ZMOM1031:
I have some code changes based on the 2.14 version
This is what I've uploaded a test version can be used
https://github.com/ZMOM1031/selfoss/releases/download/2.14-1/selfoss-2.14-1.zip
baidushurufa_2016-1-20_3-3-15

SSilence:
Thanks a lot for the PR. Can you recreate this PR for the current head?

PR #747, opened 2016-02-28
Added an indicator of "below the fold" unread entries

fmunch:
Unread entries might appear far down in the list of entries for many reasons. For example, when marking a source as read, the recent entries list will be a mix of read / unread entries.
When marking everything as read we might have missed a few unread entries because they were too far down in the list.

This PR adds an indicator displaying the number of unread entries below the fold:

Indicator of below the fold unread entries

It only appears on desktop browsers (width > 1024px), and requires the `floating_unread_count_enabled=true` config parameter to be shown (disabled by default).

It might display `?` when the number of unread entries is unknown, for example when using the search feature.

### DIFF
--- a/public/css/style.css
+++ b/public/css/style.css
@@ -381,6 +381,36 @@ body * {
         cursor:pointer;
     }
     
+/* unread count below the fold */
+
+#floating-unread {
+    display: none;
+    width: 40px;
+    position: fixed;
+    left: 185px;
+    bottom: 5px;
+    text-align: center;
+}
+
+#floating-unread .floating-unread-count {
+    display: inline;
+    min-width: 8px;
+    padding: 0 5px 0 5px;
+    background-color: #E74C3C;
+    border-radius: 30px;
+    font-size: 0.6em;
+    color: white;
+}
+
+#floating-unread .bottom-triangle {
+    position: relative;
+    bottom: 3px;
+    margin: 0 15px;
+    border-left: 5px solid transparent;
+    border-right: 5px solid transparent;
+    border-top: 5px solid #E74C3C;
+}
+
 /* content */
 
 #content {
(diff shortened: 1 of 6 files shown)


jtojnar:
Looks great. Just few comments:

* Could you rebase it to get rid of the “temporary code” commit?
* Could you add the config option into defaults.ini? (It is to make the default value more obvious.)
* Other boolean options use `0`/`1`, could you use this convention?
* Could you also update documentation (_docs/website/index.html)?
* If the tag/source has no unread items, the widget displays the number of all unread items from all the categories. The same happens on the Settings page.
* Why display the question mark and not hide the widget altogether when the number of unread is unknown?
* Could you add `aria-hidden="true"` attribute to the widget? The number might confuse screen reader users.

niol:
Interesting. Why a config option? Who wouldn't want this?

fmunch:
I fixed the issue when the tag/source has no unread items, and added the `aria-hidden` attribute, I will try to push that later today.
The config option is not really useful, I added it just to be sure that it could be disabled in case of a bug or something. I will remove it.
I added the question mark to clearly show that there _might_ be unread entries, which is different from when we _know_ that everything is read after that. But I can remove it if it is not self-explanatory.

fmunch:
Ok, so
* The config option is removed
* I added the `aria-hidden` attribute
* The bug when the tag / source had no unread entry should be fixed
* The unread count on the settings view should now be hidden
* Everything was rebased in a single commit

It can still show a question mark on a few views, should I remove this behavior too?

SSilence:
Thanks for this pr. A really nice feature. Can you fix the conflicts? Then I will merge this.

Thanks a lot!!

fmunch:
Should be fine now.

PR #748, opened 2016-02-28
Fixed the "active" flag in the sources list

fmunch:
Since f02c300, when selecting a source for the first time, the filter is applied as expected on the entries list but the highlight on the source is added and instantly removed.
After the first source selection, clicking on another source highlights the source as expected.

It affects #747 since it uses the active source to count the number of unread entries.


jtojnar:
:+1:

PR #749, opened 2016-02-28
Fixed the search offset

fmunch:
When scrolling enough to trigger the auto-stream and then going back to top to perform a search, the search keeps the previous offset instead of starting with an offset of zero.


jtojnar:
:+1:

Issue #731, opened 2016-01-04
php7 throws an error when floIcon runs, halting feed updates


tylershelton:
Hey,

Issue occurred on selfoss 2.12; upgrading to 2.14 yields same result.

Error output below with `DEBUG=1`. Occurs when refreshing via the web UI, or by hand with `cliupdate.php`, as below.

````
> php ./cliupdate.php

Methods with the same name as their class will not be constructors in a future
version of PHP; floIcon has a deprecated constructor
- helpers/Image.php:120 spl_autoload_call()
- helpers/Image.php:69 helpers\Image->loadImage()
- spouts/rss/feed.php:256 helpers\Image->fetchFavicon()
- helpers/ContentLoader.php:160 spouts\rss\feed->getIcon()
- helpers/ContentLoader.php:46 helpers\ContentLoader->fetch()
- cliupdate.php:9 helpers\ContentLoader->update()
an error occured: Methods with the same name as their class will not be constructors
in a future version of PHP; floIcon has a deprecated constructor
````
The output _feels_ informational, like a warning, but it stops the program in its tracks like an error.

Thanks in advance for taking a look!

**Comment by rookierks** [comment]
I'm also seeing this problem after the update to php7 on Arch Linux ARM. This problem is present on 2.14 and also the current git version.

This seems to be triggered only by some of my feeds. The problematic feeds are not updated but the ones that don't trigger this problem are still updated when using a cron job with cliupdate.php.

See: http://selfoss.aditu.de/forum/index.php?id=1148

**Comment by bartbes** [comment]
I've renamed the constructor as the warning suggests, and indeed now it seems to be updating fine again.

```diff
--- libs/floIcon.php.orig	2016-01-08 15:20:54.090328448 +0100
+++ libs/floIcon.php	2016-01-08 15:21:35.983869641 +0100
@@ -225,7 +225,7 @@
 	 */
 	var $images; // Array of floIconImage objects.
 	var $updated = false;
-	function floIcon() {
+	function __construct() {
 		$this->images = array();
 	}
 	function countImages() {
@@ -823,7 +823,7 @@
 	function getEntry() {
 		return $this->_entry;
 	}
-	function floIconImage() {
+	function __construct() {
 	}
 	function getHeaderSize() {
 		return strlen($this->_headerIconFormat);
```

silentboatman:
I can reproduce this bug after adding the rss-feed http://hackaday.com/feed/ and running 
```$ php ./cliupdate.php```

[...]

PR #751, opened 2016-03-19
Rename floIcon constructors

wclear:
Constructors with the same name as the class have been deprecated in PHP 7. This caused an error on a fresh installation of selfoss in an environment with PHP 7.


wclear:
Thanks! I think this solves issue [731](https://github.com/SSilence/selfoss/issues/731)

PR #754, opened 2016-04-14
Use a more secure password hash

bertptrs:
This patch changes the password hashing system to use `password_hash` and `password_verify` rather than sha512. This system uses a more expensive hashing function to make brute-forcing infeasible. It also generates its own salt securely, so you don't need to worry about that.

Unfortunately, this does bump the PHP requirement to 5.5. Since 5.3 is not officially supported anymore, this should not be too much of a problem.

### DIFF
--- a/controllers/Index.php
+++ b/controllers/Index.php
@@ -101,7 +101,7 @@ class Index extends BaseController {
         $this->view = new \helpers\View();
         $this->view->password = true;
         if(isset($_POST['password']))
-            $this->view->hash = hash("sha512", \F3::get('salt') . $_POST['password']);
+            $this->view->hash = password_hash($_POST['password'], PASSWORD_DEFAULT);
         echo $this->view->render('templates/login.phtml');
     }
     
--- a/defaults.ini
+++ b/defaults.ini
@@ -15,7 +15,6 @@ items_lifetime=30
 base_url=
 username=
 password=
-salt=lkjl1289
 public=
 html_title=selfoss
 rss_title=selfoss feed
--- a/helpers/Authentication.php
+++ b/helpers/Authentication.php
@@ -98,7 +98,7 @@ class Authentication {
     public function login($username, $password) {
         if($this->enabled()) {
             if(
-                $username == \F3::get('username') &&  hash("sha512", \F3::get('salt') . $password) == \F3::get('password')
+                $username == \F3::get('username') && password_verify($password, \F3::get('password'))
             ) {
                 $this->loggedin = true;
                 $_SESSION['loggedin'] = true;


PhrozenByte:
You can use https://github.com/ircmaxell/password_compat to keep Selfoss running with >= PHP 5.3.7

jtojnar:
Looks good to me, personally I wouldn’t mind dropping 5.3.

Could you please remove the salt option from [documentation](https://github.com/SSilence/selfoss/blob/master/_docs/website/index.html) too?

PhrozenByte:
@SSilence: Docs still need an update - the `salt` option doesn't exist anymore and Selfoss now requires PHP >= 5.5.

:-1: merging this. Dropping support of PHP 5.3 without any need (this is possible without breaking BC) is very bad :disappointed:

SSilence:
@PhrozenByte you are right I reverted this pr. SHA 512 is secure and all users with password have to configure their password protection again. And leaving PHP 5.3 support is possibly on some hoster a problem.

bertptrs:
@SSilence would you reconsider this if I solved the reconfiguration issue? PHP 5.3 *really isn't safe to use anymore*, so I would think bumping the requirement to be a good thing.

PhrozenByte:
> PHP 5.3 really isn't safe to use anymore, so I would think bumping the requirement to be a good thing.

@bertptrs: Did you even look at what https://github.com/ircmaxell/password_compat is? Did you notice **who** wrote it? Do you really know what you're talking about?

Spy-Seth:
The question is not if the `password_hash` polyfill is safe or not, but is the language is safe. PHP 5.3 has reach end of life (http://php.net/supported-versions.php) so the answer is: this version of the language is not safe any more.

Upgrading the requirement allow the project to move forward. If people want a Selfoss  that run on PHP 5.3 they could use the proper old release. And in the meantime the project could all the new feature of the language (the `password_hash` is a really good thing) without the maintenance nightmare.

But it's still a BC break and that will force a new major release. It could be interesting to look if they are other BC break subjects/features that could be done in the same release.

bertptrs:
Like @Spy-Seth, the PHP version 5.3 is the unsafe part, not its lack of `password_hash` function, which could potentially be mitigated using the link you suggested. The PHP version has been unsupported for nearly two years and currently has its fair share of unpatched exploits.

Regarding the second part, I agree that it is probably better to make this a major release thing.

PhrozenByte:
You're simply wrong @bertptrs, @Spy-Seth. PHP 5.3 still receives security updates on many long-term Linux distributions and still has a enormous market share, especially in managed environments ("hosters"). As I already said, the reason why I dislike this change is because it drops support **without any need**. If there had be a good reason to drop support of PHP 5.3, I would have said nothing. You're welcome to open PRs with great new features that require one of PHP's later language features, but at this point it's just unnecessary.

bertptrs:
Please stop saying that SHA 512 is just as secure as `password_hash`. SHA 512 can be brute forced, even salted, because it is a cheap hash to compute. `password_hash` on the other hand is slow by design, taking about a second to compute if calibrated correctly.

The reason that I made this patch in the first place is that I do not want to store a password hashed with SHA 512 anywhere, if there is a chance of it leaking.

To summarize what I have said so far:

The upsides:
* Using `password_hash` is more secure than SHA 512, mostly because it is harder to compute.
* As an added bonus, it uses a securely random generated hash, which is better than the current implementation which trusts the user to change the hash themselves.

Then for the downsides:
* This does require users to re-configure their passwords. This can be mitigated, by checking the type of the hash stored. I could write a patch for this.
* It also bumps the PHP requirement to 5.5 This is, in my opinion, a good thing, as neither 5.3 nor 5.4 is still officially supported. There may be patched versions on some shared hosting providers, but most also provide more recent PHP versions. See http://phpversions.info/ about this.

I would really like to see this feature in a future version of selfoss, but I understand your concern about BC breaks. Because of this, I would suggest to including this in the next major release.

PhrozenByte:
There are no known attacks against SHA512, neither theoretical nor practical. Brute force attacks against a salted SHA512 hash are utopian. **You can't break the SHA512 hash**. You can break the password. But not the hash. If you can break the password, the used hash algorithm makes no difference. The Blowfish cost - whose objective is to make brute force attacks against the password (not the hash!) harder - is usually set to a target computation time around 0.2s, not 1s - but your PR doesn't calculate a Blowfish cost, it just uses PHP's default, thus the computation time varies heavily depending on the hardware running on (on my laptop it's e.g. far below 0.1s). Blowfish cost furthermore is not more than calculating more rounds - what is possible with SHA512, too. The difference between `password_verify` and a simple comparison is protection against timing attacks. Implementing this without `password_verify` is a one-liner. Let me say this without wishing to offend you, but you don't seem to **really** know what you're talking about.

The official EOL of PHP isn't relevant at all. With the vast majority of Linux distributions you never received official patches anyway, even with a currently supported PHP 5.6 you don't receive official patches. That's how Linux distributions work. PHP 5.3 is still supported on many long-term Linux distributions like Ubuntu, CentOS/RHEL and some more. That's the reason why PHP 5.3 still has a large market share.

Furthermore, don't let us forget that you **can** use `password_hash` with PHP 5.3. `password_hash` is no argument to drop support of PHP 5.3 at all.

Anyway, we exchanged our arguments, now it's up to @SSilence to make a decision.

Wicloz:
@PhrozenByte, your comment is as incoherent as a paper from SCIgen, would you mind elaborating? I'm mostly confused by the sentence:
> You can't break the SHA512 hash. You can break the password. But not the hash. If you can break the password, the used hash algorithm makes no difference.

bertptrs:
@PhrozenByte please remove the ad hominem from your comment, it adds nothing to the discussion.

Regardless, an 8 character password hashed with salted SHA512 can be brute forced just fine when the salt is known. It is valid to assume that the salt leaks with the password, as they are located in the same config file. For the entire story please see this stackexchange answer: http://security.stackexchange.com/a/4691

Regarding the lack of a cost parameter, you are right, I should add it with a sensible default. But with the default settings, my crappy laptop can either take 100 guesses at a bcrypt hashed password, or 10,000,000 guesses at a sha512 hash, using the PHP implementation which is not that fast.

To test this, I used:

```
$ time php -r '$a = password_hash("apassword", PASSWORD_DEFAULT); for ($i = 0; $i < 100; $i++) password_verify($i, $a);'
php -r   8.04s user 0.03s system 98% cpu 8.209 total
$ time php -r 'for ($i = 0; $i < 10000000; $i++) hash("sha512", $i);'                                                   
php -r 'for ($i = 0; $i < 10000000; $i++) hash("sha512", $i);'  11.58s user 0.01s system 99% cpu 11.587 total
```

Also, the difference between `password_verify` and a string comparison is *not* a protection against timing attacks, but instead it extracts the cost, salt, and hash function from the stored password hash (created by `password_hash`) in order to correctly apply the hash function to the password. Please do not make false claims about what library functions do.

As a side-note, timing attacks for hash comparisons are meaningless, as having guessed part of the hash correctly tells you nothing about whether you have part of the string correctly, for any good hashing function.

PhrozenByte:
About what are we even discussing? That comparing chalk and cheese (one round sha512 vs. bcrypt with cost 12) is nonsense? That timing attacks are always an attack vector - even when comparing salted hashes? The problem of these topics is, that they are completely unrelated to this PR. Are we maybe discussing about the fact, that your claim, that PHP 5.3 "has its fair share of unpatched exploits", is wrong? Yeah, more likely.

You claim that sha512 is insecure, what simply is wrong. The way Selfoss uses sha512 is insecure, but I never claimed that the current implementation should stay the way it is. The only thing I was claiming is, that this is no reason to drop support of PHP 5.3. Heck, I mentioned https://github.com/ircmaxell/password_compat **in my first comment!** :unamused:

PR #757, opened 2016-04-21
Icons did not show up in lighttpd server

JFreakDK:
I don't know if this is a lighttpd specific issue.

With the rewrite rules described here https://github.com/SSilence/selfoss/wiki/Lighttpd-configuration the icons are not returned when pointing to /public/images/search.png. They need to point to /images/search.png

### DIFF
--- a/templates/home.phtml
+++ b/templates/home.phtml
@@ -146,8 +146,8 @@
     <!-- search -->
     <div id="search" role="search">
         <input type="search" id="search-term">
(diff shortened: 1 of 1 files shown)


niol:
This breaks apache2 installation so this cannot be merged.

Can you try to fix lighttpd rewrite rule instead by matching what's in `.htaccess` ?

SSilence:
Thanks for the pr. As niol said, this has to be solved in the rewrite rules.

PR #758, opened 2016-04-27
Delete selfoss-2.15.zip

jtojnar:


jtojnar:
It would be nice if this could be done retrospectively.

PhrozenByte:
@jtojnar, this doesn't change anything regarding the repo's size - `selfoss-2.15.zip` is still in the history. @SSilence (you're the only one who can do that because you're the only one who can force push directly to `master`), you have to take the following steps to *really* remove the file from history: https://help.github.com/articles/remove-sensitive-data/

jtojnar:
@PhrozenByte yes, I meant for @SSilence to fix it up onto ef1a1e8a8faffde396117ebd6f015241ae76ba06 and then force push.

PR #760, opened 2016-04-27
Allow to collapse/expand filter list

gerundt:
This patch allow to collapse/expand the filter list like the tags and the sources.

### DIFF
--- a/public/css/style.css
+++ b/public/css/style.css
@@ -87,7 +87,23 @@ body * {
         #nav-mark:hover {
             color:#000000;
         }
+
+    #nav-filter-wrapper  {
+        overflow:hidden;
+    }
     
+        .nav-filter-collapsed {
+            cursor:pointer;
+            padding-left: 20px;
+            background: url(images/nav-collapsed.png) center left no-repeat;
+        }
+        
+        .nav-filter-expanded {
+            cursor:pointer;
+            padding-left: 20px;
+            background: url(images/nav-expanded.png) center left no-repeat;
+        }
+
     #nav-filter-title {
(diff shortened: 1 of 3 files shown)


jtojnar:
Looks nice but the utilised area shrinks as the filters are collapsed making the whole thing rather useless.

expand

gerundt:
Mhh, I tested it with Firefox, IE 11 and Android Chrome and have no problems. Which browser use you?

jtojnar:
I tried Firefox and Midori, both exhibit the same behaviour:

https://gfycat.com/ThriftySomberIrishwaterspaniel

gerundt:
Ok, I see the problem now too. I will look to find a solution. Thank you for your hint!

gerundt:
I hope 0873875 fix the problem. Can you try it? :)

jtojnar:
This is more usable but the transition looks horrible. How about moving the filters inside the container with scrollbar?

https://gfycat.com/HomelyWelllitHorsechestnutleafminer

gerundt:
Yes, moving the filters inside the container with the scrollbar could be an option! But I am not sure, if I have enough JavaScript and CSS knowledge for this. :-(

jtojnar:
Hi @gerundt. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contribution under `GPL-3-or-later`?

gerundt:
Hi, i am fine with GPL 3 or later for all my commits!

PR #765, opened 2016-05-16
Update style.css

Poorchop:
Wrap text in code blocks to prevent column overflow

It looks like my proposed rule can cause some issues as well. In the example in my screenshot, the line numbers are a separate element, which causes a mismatch in the middle column, but not in the left column for some reason. I'm just throwing this out there because as it stands, the text spilling over into other columns is much worse in my opinion.

untitled

Top is before, bottom is after.

### DIFF
--- a/public/css/style.css
+++ b/public/css/style.css
@@ -572,6 +572,10 @@ body * {
         .entry-content blockquote {
             overflow-x: unset;
         }
+        
+        .entry-content pre {
+            white-space: pre-wrap;
+        }
     
         .entry-content p,
         .entry-content ol,


Poorchop:
The best solution would be to somehow implement `overflow-x: scroll` but there is no special class or ID in order to identify code blocks. At least in this case, it's just a regular `div` immediately after `.entry-content`.

Poorchop:
This would be better solved with JavaScript.

bendem:
I'm not sure about that. Also, why not just `code, pre { overflow-x: auto; }`? Maybe you can even apply it directly to the column?

Poorchop:
I had a good picture saved that would have answered your question very clearly, but I deleted it and my server is down so I can't access my feed. Setting `overflow-x: auto` or even `scroll` didn't work because the `pre` tag is nested deep within a `div` that needs to have the style applied to it. It was something along the lines of `.entry-content > div > div > div > pre` or along those lines. In my testing, the `div` immediately following `.entry-content` needed to have the overflow style applied to it. That's the only way that it worked.

Since that `div` in question doesn't have a class, ID, or any unique identifier such as a property, I think that it needs to be targeted with JavaScript. I think you would need to target `[.entry-content].firstChild` and apply `{ overflow-x: auto; }` there.

I was just going to rebase and submit that but like I said, I'm having server issues and I didn't want this to get merged in the meantime.

PR #768, opened 2016-05-26
Apply upstream patch to f3/base.php

### LINKED ISSUE #767: Illegal parameter name in lambda
In the `base.php` of the f3 library, there is the following code:

``` php
        /**
        *       Redirect a route to another URL
        *       @return NULL
        *       @param $pattern string|array
        *       @param $url string
        */
        function redirect($pattern,$url) {
                if (is_array($pattern)) {
                        foreach ($pattern as $item)
                                $this->redirect($item,$url);
                        return;
                }
                $this->route($pattern,function($this) use ($url) {
                        $this->reroute($url);
                });
        }
```

This code uses the name `$this` as an argument name for the lambda. After upgrading PHP to 7.0.7 (Arch Linux) this produces a fatal error, as `$this` is not a legal argument name. I worked around it by renaming it `$that`, which appears to work, but someone more familiar with the F3 framework should probably see if that is a good idea. My version now looks like this:

``` php
                $this->route($pattern,function($that) use ($url) {
                        $that->reroute($url);
                });
```

bertptrs:
Fix #767. Now uses the parameter name `$fw` instead of `$this`, which
does not produce a fatal error. This comes from version 3.5.0 upstream.

### DIFF
--- a/libs/f3/base.php
+++ b/libs/f3/base.php
@@ -1182,8 +1182,8 @@ class Base extends Prefab {
 				$this->redirect($item,$url);
 			return;
 		}
-		$this->route($pattern,function($this) use ($url) {
-			$this->reroute($url);
+		$this->route($pattern,function($fw) use ($url) {
+			$fw->reroute($url);
 		});
 	}
 


jernst:
Just registering my interest in seeing this merged. We currently do this patch ourselves in UBOS, but it would be nice if we didn't have to.

jtojnar:
This can now be closed as F3 was upgraded.

PR #769, opened 2016-05-26
remove $this in anonymous function for PHP7 compatibility

amiga23:
in PHP7
function($this) use ($url) {
leads to PHP Fatal error:  Cannot use $this as parameter in /selfoss/libs/f3/base.php on line 1185


amiga23:
Sorry, did not see that this is already fixed upstream

Issue #740, opened 2016-02-04
Empty response when fetching 'tags' through API


f3bruary:
When I browse to myselfossinstance.com/tags it returns [ ]. In the logs I see: "Error Unimplemented method for sqlite: numberOfUnreadForTag", but I don't know if it's related.

niol:
``numberOfUnreadForTag`` was removed a while ago. Please check your installation and the version you are using.

f3bruary:
I updated 10 minutes ago. The only thing I didn't replace during the update was the data dir

jtojnar:
@niol [not completely](https://github.com/SSilence/selfoss/blob/master/controllers/Items.php#L149), apparently.

PhrozenByte:
Introduced with #714 (this feature doesn't work for me anyway...)

niol:
OK sorry I'm lagging behind git HEAD.

PhrozenByte:
No, you're completely right, [`\daos\Items::numberOfUnreadForTag()` has been removed](https://github.com/SSilence/selfoss/search?utf8=%E2%9C%93&q=numberOfUnreadForTag), but #714 still uses it and introduces this bug. I'm not using the current HEAD either.

niol:
This can be closed.

PR #770, opened 2016-05-27
Fix read count of secondary tags

jtojnar:
Cherry-picked https://github.com/wazari972/selfoss/commit/3853d6423c3925978119ed0be0171da1611ca5a4, should fix the error reported in #740 (not the issue itself though). It will also make hidden tags actually work as promised in #714.

Problems not fixed in this pull request:
* _Unread_ filter count is subtracted multiple times when a feed has multiple hidden tags (it can even show negative count).
* _Unread_ filter count is not AJAX-updated when you tag/untag a source with a hidden tag.
* Number of unread items below the fold (#747) doesn’t subtract the hidden items.
* Private tags are still displayed in sidebar, although they correctly don’t show any content.
* There is not documentation.

### DIFF
--- a/controllers/Index.php
+++ b/controllers/Index.php
@@ -50,14 +50,12 @@ class Index extends BaseController {
         $this->view->statsUnread = $stats['unread'];
         $this->view->statsStarred = $stats['starred'];
 
-        if ($tagsDao->hasTag("#")) {
-		foreach ($tags as $tag) {
-			if (strcmp($tag["tag"], "#") !== 0) {
-				continue;
-			}
-			$this->view->statsUnread -= $tag["unread"];
-		}
-	}
+        foreach ($tags as $tag) {
+            if (strpos($tag['tag'], '#') !== 0) {
+                continue;
+            }
+            $this->view->statsUnread -= $tag['unread'];
+        }
 
         // prepare tags display list
         $tagsController = new \controllers\Tags();
--- a/controllers/Items.php
+++ b/controllers/Items.php
@@ -146,17 +146,15 @@ class Items extends BaseController {
 
         $itemsDao = new \daos\Items();
         $stats = $itemsDao->stats();
-	$stats['unread'] -= $itemsDao->numberOfUnreadForTag("#");
 
-	$tagsDao = new \daos\Tags();
-	$tags = $tagsDao->getWithUnread();
-        if ($tagsDao->hasTag("#")) {
-            foreach ($tags as $tag) {
-                if (strcmp($tag["tag"], "#") !== 0) {
-                    continue;
-                }
-                $stats['unread'] -= $tag["unread"];
+        $tagsDao = new \daos\Tags();
+        $tags = $tagsDao->getWithUnread();
+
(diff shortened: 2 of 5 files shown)


Poorchop:
Not that it makes a difference, but isn't the indentation off here? Why 8 spaces instead of 4?

jtojnar:
Actually there were even tabs mixed in, I just took the commit from @wazari972’s fork and opened a PR.

I edited the commit so it should match the “coding style” better now.

wazari972:
sorry for the indentation problems, tab+spaces is emacs style :$ I don't know how to tell it to follow the original style !

I don't remember exactly what was the problem solved by that patch, but I use it since I wrote it and secondary tags count work well.

(the only minor problem on the JS side is the -1 on the *main* tag count that should not happened, but I'm too lazy to fix it :- )

jtojnar:
The original code only considered hidden **only** the tag '#'. This change makes it so any tag **containing** '#' is hidden.

I only submitted this pull request to remove `numberOfUnreadForTag` call but making #714 work is a nice bonus.

The negative unread count can happen when a feed has multiple hidden tags (its items are subtracted multiple times). Another (although minor) problem is that the filter (unread) count is not AJAX-updated when you tag/untag a source with a hidden tag. And of course, the missing documentation is also pain. But these can be fixed in a separate PR.

PhrozenByte:
Instead of matching a `@` and `#` anywhere in the string, it would be great to check just the first char - this should avoid accidental matchings.

jtojnar:
@PhrozenByte Yes, I was not sure whether this was intentional. Maybe @wazari972 accidentally used `strcmp` instead of `strpos` and changed it to `strstr` when it did not work.

Now it should only hide tags beginning with '#'.

PhrozenByte:
Great, thanks @jtojnar! Can you please implement this for `@`, too?

jtojnar:
@PhrozenByte this should probably do it.

It is little messy because selfoss stores tags a text. Maybe the `foreach`es could be moved to a separate function, I am not sure where should it go in this though. IMHO this thing whole thing should be handled inside the controllers not in DAOs (if anywhere).

jtojnar:
Also this doesn’t really work well with #747

wazari972:
@jtojnar @PhrozenByte 
> Instead of matching a @ and # anywhere in the string, it would be great to check just the first char - this should avoid accidental matchings.
> Yes, I was not sure whether this was intentional. Maybe @wazari972 accidentally used strcmp instead of strpos and changed it to strstr when it did not work.

if I remember correctly, the problem was that the tags are stored in one string, like "tag, #hidentag, tag, #hidentag...". So checking that # is at position 0 doesn't work. But I can't remember why I didn't look for " #..." instead of deciding that "# can be anywhere in the name" ... (I'm not good at PHP and don't know it's API very well, maybe that's the answer ;-)

jtojnar:
@wazari972 Yes, after trying to change it for `@`, I realised that was probably your reason for not checking position 0. Anyway, in 7bc88f9 it should be handled correctly.

PR #772, opened 2016-06-01
Fix unclosed link tags

Poorchop:
Failing to explicitly close these tags was resulting in erratic text coloring.


jtojnar:
Maybe it would be even better to make the anchors `id`s of the headings instead. Empty elements feel a little awkward.

Poorchop:
You're probably right but I thought that this is how anchors are normally handled. I think it should be

```html
<a class="anchor" href="https://selfoss.aditu.de/#configuration">
  <h2>Configuration</h2>
</a>
```
This will actually make use of the `<a>` tag instead of eliminating it altogether and just adding an ID to the `<h2>` tag itself. This way the user can just click on a header and immediately have a link to the section instead of having to manually type in the ID. What do you think?

Poorchop:
Okay I totally messed up this rebase and I keep making it worse. Hopefully I can figure this out but it is becoming catastrophic.

Issue #589, opened 2014-10-19
Multi-user support


auzias:
I really like your workand would like to improve it [more](https://github.com/SSilence/selfoss/pull/585).
I got two ideas (which are two needs too), I would like to know which one the community would rather have:
1. Firefox App,
2. Multi-users support.

Btw, a twitter account for selfoss might be handful sometimes ;)
Have a nice day!

jtojnar:
Closing in favour of #882 since this is quite broad.

PR #777, opened 2016-06-15
Add feature 'Remember Me'

Gugiman:
Added a Feature to stay logged in.

Very Useful for mobile User with website bookmark.

### DIFF
--- a/helpers/Authentication.php
+++ b/helpers/Authentication.php
@@ -31,6 +31,8 @@ class Authentication {
         $cookie_secure = isset($_SERVER['HTTPS']) && !empty($_SERVER['HTTPS']) && $_SERVER['HTTPS']!=="off";
         $cookie_httponly = true;
 
+		$remainlogged = false;
+		
         // check for SSL proxy and special cookie options
         if(isset($_SERVER['HTTP_X_FORWARDED_SERVER']) && isset($_SERVER['HTTP_X_FORWARDED_HOST'])
            && ($_SERVER['HTTP_X_FORWARDED_SERVER']===$_SERVER['HTTP_X_FORWARDED_HOST'])) {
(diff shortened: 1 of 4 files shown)


bendem:
Just checking, this means I can bypass login on any selfoss install using this PR by setting `remainlogged=1` as a cookie, right?

Gugiman:
Yep absolutely. My bad. Sorry for that. 

What about assigning the password hash + salt as the cookie value? Since users aren't stored in database it's hard to save a token which is new generated every (skipped) login.

jtojnar:
The problem with this is that if you allow to authenticate using the hash of the password, you are essentially making the hash a password. It would be exactly the same as storing the unhashed password in a cookie. If one device is compromised, you have to change the password.

The only safe way to get the session to last is by generating a session key and storing it in on the client – which is exactly what PHP is doing with its `$_SESSION`, only that by default the expiration is somewhat shorter than we would like.

If you are unable to change the duration of the `$_SESSION` cookie, you will have to replicate this whole mechanism yourself, including sending a unique token to each device.

Now, there is a way to do this without a server-side storage but it requires the use of cryptography (IIRC Json Web Token uses this method) which I would not trust myself to implement correctly.

Disregarding this method, it is simply not possible to do this securely without a database. Maybe you could do this and finally implement #589.

PhrozenByte:
You can increase PHP's session lifetime with `session.gc_maxlifetime` in your `php.ini`. However, please note that some distributions (e.g. Debian) don't use PHP's built-in garbage collector, but a cronjob to remove old session files. If your intention with this PR is to prevent accidental logouts while the page is still open (very annoying), you may want to use https://github.com/SSilence/selfoss/pull/611 for session keep alive.

Gugiman:
Intresting. I may give it a try.

Bigger changes to provide a rembered login would may be the beginning of multiuser support.

Maybe one day i have some time to look at it. But this PR seems not to be the most secure solution so I close it for now.

PR #781, opened 2016-06-22
Optionaly proxify images src with atmos/camo

luxifer:
I used selfoss every day with a Let's Encrypt certificate but the connection is not fully encrypted when some feeds have images that does not come from https. To address this situation I used https://github.com/atmos/camo (an http proxy to serve images through SSL). It's optional so if the configuration for host and key are empty, nothing is done.

### DIFF
--- a/_docs/website/index.html
+++ b/_docs/website/index.html
@@ -2,18 +2,18 @@
 <html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en">
 <head>
     <title>selfoss - the open source web based rss reader and multi source mashup aggregator</title>
-    
+
     <meta http-equiv="content-type" content="text/html; charset=utf-8" />
     <meta http-equiv="language" content="en" />
     <meta name="copyright" content="Tobias Zeising" />
     <meta name="keywords" content="selfoss rss reader webbased mashup aggregator tobias zeising aditu" />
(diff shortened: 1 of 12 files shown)


niol:
This looks very useful, thanks.

But is seems to be a lot of code for something that could be implemented using a htmLawed hook and some f3 standard functions.

luxifer:
I'll check this out and rewrite my PR to be shorter and leaner. This was a
proof of concept. Thanks for the insight.

Florent Viel

jtojnar:
I like it the way it is implemented here, separation of concerns-wise.

jtojnar:
Hi @luxifer. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contribution under `GPL-3-or-later`?

luxifer:
I'm fine with GPL-3 licensing

jtojnar:
@luxifer Unfortunately, the fact that GPL 3 is ambiguous (as described in https://gnu.org/licenses/identify-licenses-clearly.html) is the whole reason for the uncertainty. Could you please clarify whether you are okay with `GPL-3-or-later` rather than `GPL-3-only` specifically?

luxifer:
Sorry for the confusion but I'm fine with any licensing ;)

PR #783, opened 2016-06-25
Fix img src path to search.png and remove.png

rpe-github:
The img src to remove.png and search.png (added with 462e91ccfb522b37a20a848d57d77d15364709af) causes broken links to these images.

image

default.log:
06-25-16 9:26:23 Error HTTP 404 (GET /public/images/remove.png)
06-25-16 9:26:23 Error HTTP 404 (GET /public/images/search.png)

### DIFF
--- a/templates/home.phtml
+++ b/templates/home.phtml
@@ -148,8 +148,8 @@
     <!-- search -->
     <div id="search" role="search">
         <input type="search" id="search-term">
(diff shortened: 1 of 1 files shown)


jtojnar:
Sorry about that, probably a remnant of when I was playing with `.htaccess`.

PR #787, opened 2016-07-06
Added Slovak translation

imdyske:


jtojnar:
Shouldn’t “před” be “pred”?

Mychal713:
Ah, thanks. Totally missed that one.

jtojnar:
Thanks, can you squash it into one commit?

Mychal713:
Sorry I'm new to git. I have no idea how to squash them together.

jtojnar:
That’s all right, this project does not really require a strict coding style. If you want, you can issue `git rebase -i HEAD~2` which will start interactive manipulation with last two commits. Then you can change the second `pick` command to `f` (fixup) and exit the editor. After that you will have to force the changes to github using `git push --force`.

Mychal713:
It's giving me `error: could not apply bd8d286... Added Slovak translation` then I abort the rebase, but no matter which commit I choose to fixup the error is the same.

Mychal713:
Ok, I have managed to resolve the problem. Thanks for your help.

jtojnar:
Hi @imdyske, previously it was not clear if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contribution under `GPL-3-or-later`?

PR #792, opened 2016-08-08
Don't delete config.ini during update

Sailing74:


jtojnar:
Hi @Sailing74. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions under `GPL-3-or-later`?

Sailing74:
Yes, perfectly fine by me.

PR #807, opened 2016-10-02
Use https for Heise RSS Feeds

DeX77:
Heise recently enabled https access for all services. So use it.


jtojnar:
Duplicate of #804

Issue #534, opened 2014-07-08
Internal Server Error (UPDATE)


LabSen:
When I try to update my feeds, I have an Internal Server Error:

Internal Server Error

PDOStatement: Incorrect string value: '\xF0\x9F\x92\x8B' for column 'title' at row 1

debug log:

07-08-14 10:05:58 Debug ---
07-08-14 10:05:58 Debug start fetching source "Klicktoris (id: 107) 
07-08-14 10:05:58 Debug spout successfully loaded: spouts\rss\feed
07-08-14 10:05:58 Debug fetch content
07-08-14 10:05:58 Debug minimum date: 2013-07-08 10:05:58
07-08-14 10:05:58 Debug start item fetching
07-08-14 10:05:58 Debug start insertion of new item "babygirlssweetsurrender:

💋"
07-08-14 10:05:58 Debug item content sanitized
07-08-14 10:05:59 Debug icon generated: http://31.media.tumblr.com/avatar_f2f132a9919f_128.png

When i delet the source klicktoris works perfect. Probably a charset problem

vegbrasil:
Hi guys, I'm having the same problem.

It happens with Twitter spout trying to fetch a tweet from `@htcvive`.

```
01-17-16 14:18:14 Debug start fetching source "@htcvive (id: 16) 
01-17-16 14:18:14 Debug spout successfully loaded: spouts\twitter\usertimeline
01-17-16 14:18:14 Debug fetch content
01-17-16 14:18:14 Debug minimum date: 2015-12-18 14:18:14
01-17-16 14:18:14 Debug start item fetching
01-17-16 14:18:14 Debug start insertion of new item "Engadget (RT HTC Vive):<br>Editor-in-Chief @Numeson demoing @htc Vive. {Follow us on Snapchat for more: gadgetsnaps 👻} #virtualreality <a href="https://t.co/IIIlEyoMeo" target="_blank">https://t.co/IIIlEyoMeo</a>"
01-17-16 14:18:14 Debug item content sanitized
01-17-16 14:18:14 Debug thumbnail generated: http://pbs.twimg.com/ext_tw_video_thumb/685938382954872833/pu/img/hQ_8ltr0Hj3JC6Y8.jpg
01-17-16 14:18:14 Debug icon generated: http://pbs.twimg.com/profile_images/655059892022022144/Pq3Q_1oU_normal.png
[...]

PR #813, opened 2016-11-03
Fix #811 - fetch and show emoji icons

### LINKED ISSUE #811: Problems with Twitter spout for accounts with emojis
For example, the tweet https://twitter.com/archiloque/status/794160285615616000 

captura de pantalla de 2016-11-03 10 26 44

appears the following way in selfoss:

captura de pantalla de 2016-11-03 10 27 26

severo:
See https://github.com/SSilence/selfoss/issues/811 and https://github.com/SSilence/selfoss/issues/534#issuecomment-172329123

### DIFF
--- a/libs/f3/db/sql.php
+++ b/libs/f3/db/sql.php
@@ -461,8 +461,7 @@ class SQL {
 		if (!$options)
 			$options=array();
 		if (isset($parts[0]) && strstr($parts[0],':',TRUE)=='mysql')
-			$options+=array(\PDO::MYSQL_ATTR_INIT_COMMAND=>'SET NAMES '.
-				strtolower(str_replace('-','',$fw->get('ENCODING'))).';');
+			$options+=array(\PDO::MYSQL_ATTR_INIT_COMMAND=>'SET NAMES utf8mb4;');
 		$this->pdo=new \PDO($dsn,$user,$pw,$options);
 		$this->engine=$this->pdo->getattribute(\PDO::ATTR_DRIVER_NAME);
 	}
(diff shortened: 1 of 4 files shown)


libs/f3/db/sql.php
```diff
@@ -461,8 +461,7 @@ function __construct($dsn,$user=NULL,$pw=NULL,array $options=NULL) {
 		if (!$options)
 			$options=array();
 		if (isset($parts[0]) && strstr($parts[0],':',TRUE)=='mysql')
-			$options+=array(\PDO::MYSQL_ATTR_INIT_COMMAND=>'SET NAMES '.
-				strtolower(str_replace('-','',$fw->get('ENCODING'))).';');
+			$options+=array(\PDO::MYSQL_ATTR_INIT_COMMAND=>'SET NAMES utf8mb4;');
```
jtojnar: This will break when we update f3 framework. Better raise issue [there](https://github.com/bcosca/fatfree/issues/).

severo: OK, thanks. I created the following issue, but I don't think I can help more on this problem.

https://github.com/bcosca/fatfree/issues/976

jtojnar:
Also migration script will probably be needed, otherwise users will receive the same error you got here https://github.com/SSilence/selfoss/issues/811#issuecomment-258187844, just with charsets reversed.

jtojnar:
@SSilence this is not ready to be merged, it will break existing installations. We need migration script first.

jtojnar:
@SSilence I can add it though.

severo:
It would be wonderful! I'm afraid I would last a lot before doing it...

jtojnar:
@SSilence I introduced the upgrade script in #819

PR #819, opened 2016-11-17
Add upgrade script for conversion to utf8mb4

jtojnar:
PR #813 updated the MySQL code to use proper unicode charset, this commit adds an upgrade script for seamless transition of existing installations.


jtojnar:
Also could I become a collaborator? Being able to add labels to issues and pull requests would be greatly beneficial.

SSilence:
Thanks a lot for the update script! I run selfoss with sqlite and haven't see this issue!

jtojnar:
@SSilence I mentioned it in one of the comments. That’s one of the reasons why I would like to be a collaborator – I could add ‘needs changes’ label to the PR to make it more clear. I could also close some issues that are duplicates or no longer relevant, there are quite a lot of them in the tracker.

PR #826, opened 2016-12-17
Fix encoding problems caused by camo

lumiru:
Loading HTML in DomDocument create encoding issues.
Additionally, regex are faster than loading a document tree.

I often found this issue with fulltextrss spout and French feeds.

There might be another problem with encoding with this spout.
Indeed, I get similar problems on selfoss android app even if I apply this fix.


jtojnar:
This does not look like like it handles multiple images well. Regexes might be slightly faster but are much easier to make mistake.

lumiru:
Oh sorry, I made a mistake.

@jtojnar Please note that the used regex is the same as the one used for delayed image load.

Issue #750, opened 2016-03-07
Feature: Wait-Time for next update


Thomas--F:
Hi Folks,

every time I call the feed-update, every feed is fetched. But I have a lot of feeds, that updates only once a day or once a week (e.g. some online-cartoons) while others updates very frequently (e.g. news-feeds). It would be great, if I can set a wait-option at a source, that says:
Skip this feed during update until a given time period is expired.

There is already a "lastupdate"-column in sources. So we need a new update-period-column.
I can code the changes, but only if this feature is wanted. Changing the database shouldn't be done without a good reason.

niol:
Better, the update period could be computed from past entries: no configuration, no database change.

Thomas--F:
That wouldn't run well. You have to enable this per feed, so you must save the config somewhere.
Otherwise the update will go crazy, if a feed updates very randomly.

I implement this feature on my RasPi and it works fine so far. I added a column to the database where you can save the minimum wait time in seconds, until this feed is fetched again.
I have some feeds on "0 seconds", so the update every time, the cronjob runs (in my case every 10 minitues). Others run on 3600 seconds (once an hour) or even 86400 seconds (once a day).

The workload on the RaspberryPie has dropped a lot after the changes.

niol:
I don't see your point about the update going _crazy_ and I doubt many people will take the time to manualy configure this per feed.

martinsauter:
I think it should really be configurable per feed as I have feeds for example that update only once a day or only once every 2 or 3 days. Nevertheless I still want Selfoss to look for an update during every run because I want to find out that there is something new **IMMEDIATELY** and not only much later just because there is only a new post every couple of days. 

From my point of view the frequency with which to check for news in a feed and the frequency in which a new post appears in a feed have **nothing** to do with each other in many cases.

jtojnar:
If an average time interval between *last n items* is used it might be unreliable when items don’t follow homogeneous distribution (i.e. articles are published in the evening).

It could record a release times and compute the probability distribution of updates but that would probably be an overkill.

@Thomas--F In your implementation, is this a property of RSS spout or is it possible to configure it for any spout?

Thomas--F:
@niol The default ist "0 seconds waittime", so if you don't configure anything, everything is running the old way

[...]

PR #829, opened 2016-12-19
add waittime-parameter

Thomas--F:
please check and test the code for issue  #750

It works fine for me, but I only tested the mysql-part. The other daos are changed too and I think they're working, but I cannot be sure.

The new parameter "Waittime" at the sources skips a feet-update, if the amount of seconds in this field isn't reached after the last update.
You can still update the source manually from the sources-drop-down.


helpers/ContentLoader.php
```diff
@@ -43,7 +43,13 @@ public function __construct() {
     public function update() {
         $sourcesDao = new \daos\Sources();
         foreach($sourcesDao->getByLastUpdate() as $source) {
-            $this->fetch($source);
+        	  $last = $source['lastupdate'];
+        	  $wait = $source['waitperiod'];
+        	  if (($last + $wait) < time()){
+	            $this->fetch($source);
+//            } else {
```
jtojnar: Why is this commented code here?

daos/pgsql/Database.php
```diff
@@ -199,6 +199,13 @@ public function __construct() {
                         INSERT INTO version (version) VALUES (9);
                     ');
 				}
+                if(strnatcmp($version, "11") < 0) {
+                    \F3::get('db')->exec(array(
+                        'ALTER TABLE `' . \F3::get('db_prefix') . 'sources` ADD waitperiod INT NOT NULL DEFAULT 0;'
```
jtojnar: Missing comma after first array item.

templates/source.phtml
```diff
@@ -57,6 +57,9 @@
 
         <!-- filter -->
         <li><label for="filter-<?= $idAttr ?>"><?PHP echo \F3::get('lang_source_filter')?>:</label> <input id="filter-<?= $idAttr ?>" type="text" name="filter" value="<?PHP echo isset($this->source) ? $this->source['filter'] : '' ; ?>" placeholder="<?PHP echo F3::get('lang_source_optional'); ?>" /></li>
+
+        <!-- waitperiod -->
+        <li><label for="filter-<?= $idAttr ?>"><?PHP echo \F3::get('lang_source_waitperiod')?>:</label> <input id="filter-<?= $idAttr ?>" type="text" name="waitperiod" value="<?PHP echo isset($this->source) ? $this->source['waitperiod'] : '' ; ?>" placeholder="<?PHP echo F3::get('lang_source_seconds'); ?>" /></li>
```
Thomas--F: good point. done

helpers/ContentLoader.php
Thomas--F: If you activate these lines you get a feedback on the cliupdate when a feed is skipped. I wasn't sure if that's ok.

Thomas--F:
@jtojnar: Thanks for yout help.

About the commented lines... should I use them or delete them?

helpers/ContentLoader.php
jtojnar: It would make sense to log it, I am not sure where should the echoed output appear.

jtojnar:
I would replace the commented line with proper logging.

Additionally, it might be nice to show the last updated time or that it was skipped on the source management and/or source view page.

Thomas--F:
I changed the output to the logger (log-level INFO). The local test was fine.

jtojnar:
I tested it with other database engines. SQLite seems to work fine but PostgreSQL fails when upgrading the database to version 11. It turns out the backticks used for enclosing identifiers are [non-standard](https://stackoverflow.com/questions/10573922/what-does-the-sql-standard-say-about-usage-of-backtick) and SQLite only recognizes them [to be more compatible with MySQL](https://sqlite.org/lang_keywords.html). Double quotes should be used instead, afterwards it is good to go.

Thomas--F:
I removed the backticks and the db_prefix, because the other statements dont't use them too.

jtojnar:
Yes, that is another problem, though people who run PostgreSQL databases probably won’t need such a last-resort measures like table prefixes; that is intended mainly for extremely low cost shared hostings providing only one database.

One last thing I noticed, the English “Waittime” label should probably be two words.

Thomas--F:
Fixed it.

SSilence:
Hm I don't think this is a really necessary usecase. Traffic is a very cheap thing and I think no provider has a problem if my selfoss pulls a rss feed every 30 minutes which updates rarely.

Thomas--F:
It's ok. Your software - your rules ;-)

For me this is a very useful feature becaus I run several services on my RaspberryPie and this little things tends to slow down when receiving many feeds at the same time.
I will try to stay up-to-date with my fork whenever you release a new version, so everyone who wants this feature can download it there.

Keep up the good work! Thank you for this very useful software!

jtojnar:
@Thomas--F https://github.com/SSilence/selfoss/wiki/Maintained-feature-patches

Thomas--F:
I updated my fork to the current version (some commits after release 2.17)

Issue #774, opened 2016-06-12
Incorrectly calculated offset for loading new items


jtojnar:
When you mark some source items as read and then scroll down to load more, it loads items with the same offset as if you did not read anything, causing some items to be skipped.

![selfoss-queue](https://cloud.githubusercontent.com/assets/705123/15994007/dbe4145c-30f7-11e6-8a95-c85b78d45fbd.png)

The offset when loading new items should be decreased by the number of items marked as read.

PR #834, opened 2016-12-28
Make selfoss navigable

niol:
These patches make selfoss navigable using the url hash. This makes the url reflect whats displayed in selfoss. This also makes it possible to reload the item that was open before a browser crash or when the tab containing selfoss was offloaded from memory on a mobile browser.

### DIFF
--- a/public/js/selfoss-events-navigation.js
+++ b/public/js/selfoss-events-navigation.js
@@ -121,8 +121,6 @@ selfoss.events.navigation = function() {
                 type: 'GET',
                 success: function(data) {
                     selfoss.refreshSources(data.sources);
-                    selfoss.sourcesNavLoaded = true;
-                    toggle();
                 },
                 error: function(jqXHR, textStatus, errorThrown) {
                     selfoss.showError('Can not load nav stats: '+
--- a/public/js/selfoss-events-navigation.js
+++ b/public/js/selfoss-events-navigation.js
@@ -45,6 +45,7 @@ selfoss.events.navigation = function() {
         $(this).addClass('active');
         
         selfoss.filter.offset = 0;
+        selfoss.filter.extra_ids.length = 0;
         selfoss.reloadList();
         
         if(selfoss.isSmartphone())
@@ -74,6 +75,7 @@ selfoss.events.navigation = function() {
             selfoss.filter.tag = $(this).find('span').html();
             
         selfoss.filter.offset = 0;
+        selfoss.filter.extra_ids.length = 0;
         selfoss.reloadList();
         
         if(selfoss.isSmartphone())
@@ -99,6 +101,7 @@ selfoss.events.navigation = function() {
         selfoss.filter.source = $(this).attr('id').substr(6);
             
         selfoss.filter.offset = 0;
+        selfoss.filter.extra_ids.length = 0;
         selfoss.reloadList();
         
         if(selfoss.isSmartphone())
--- a/public/js/selfoss-events-navigation.js
+++ b/public/js/selfoss-events-navigation.js
@@ -148,16 +148,12 @@ selfoss.events.navigation = function() {
         // show
         if(nav.is(':visible')==false) {
             nav.slideDown(400, function() {
-                location.hash = "nav";
                 $(window).scrollTop(0);
             });
             
         // hide
         } else {
             nav.slideUp(400, function() {
-                if(location.hash=="#nav") {
-                    location.hash = "";
-                }
                 $(window).scrollTop(0);
             });
         }
--- a/public/js/selfoss-events-navigation.js
+++ b/public/js/selfoss-events-navigation.js
@@ -40,15 +40,16 @@ selfoss.events.navigation = function() {
             selfoss.filter.type='unread';
         else if($(this).hasClass('nav-filter-starred'))
             selfoss.filter.type='starred';
+
+        if( selfoss.events.section != selfoss.filter.type ) {
+            location.hash = '#' + selfoss.filter.type
+                            + '/' + selfoss.events.subsection;
+        }
         
         $('#nav-filter > li').removeClass('active');
         $(this).addClass('active');
         
-        selfoss.filter.offset = 0;
-        selfoss.filter.extra_ids.length = 0;
-        selfoss.reloadList();
-        
-        if(selfoss.isSmartphone())
+        if(selfoss.isSmartphone() && $('#nav').is(':visible'))
             $('#nav-mobile-settings').click();
     });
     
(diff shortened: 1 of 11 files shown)


jtojnar:
For fixing #774, would not it be cleaner to use `where id < :id_of_last_loaded_item` instead of `offset` + `extra_ids`? It could also be slightly faster as we can have index on `id`. I like the rest of 2f87dd12482ce4129c77e58015a64d58cc1d4ec4, though.

jtojnar:
Regarding the second part, I discovered some edge cases:

1. Tag names are URL-encoded twice, causing tags using Unicode not to work. For example, `新闻国内` turns into
`/selfoss/?offset=0&itemsPerPage=50&search=&type=unread&tag=%25E6%2596%25B0%25E9%2597%25BB%25E5%259B%25BD%25E5%2586%2585&source=&sourcesNav=false&ajax=true` – notice the `%25` instead of `%`.
2. Visiting item on one of the lower pages does not expand it. We might show the visited item on top, as loading all the preceding items when visiting an older item might not be feasible.
Now I also see why you choose the `extra_ids`, though I still think that using the `id < :id_of_last_loaded_item` might be better than sending the list of all the read items when loading next page.
3. Visiting source listing does not expand the sources in the sidebar like clicking the item source does.
4. Item from a source not in a certain tag can be visited through the tag’s URL. This is probably not worthy of worrying about.

I did not check the code in 8b7f8d24f306981c7afdeedadd099da9b1a9401f very deeply yet.

bendem:
Agreed with @jtojnar, here is some explanation in regard to offset/limit pagination is considered bad.
https://blog.jooq.org/2016/08/10/why-most-programmers-get-pagination-wrong/

niol:
Point 1, I cannot reproduce with multiple unicode strings using Chromium as browser.
Regarding the _cleaner_ and ```id``` comparison, this would not work as we order by ```datetime``` and not ```id``` and that may not be the same order (a source crawled after but with an older item will have a higher ```id```).  I'll run some basic tests to see if _seek_ pagination is worth it regarding performance. It looks very interesting but is not related to fixing that bug.

jtojnar:
Indeed, it seems to work in Chromium as `document.location.hash === "#unread/tag-新闻国内"`. In Firefox it is `"#unread/tag-%E6%96%B0%E9%97%BB%E5%9B%BD%E5%86%85"`.

As for the *seek* pagination, my bad. The items would have to be sorted by `datetime DESC, id DESC` and then filtered using `(datetime, id) < (:last_datetime, last_id)`; index could be created for this column pair.

Although I think the performance difference will be negligible in most installation, as selfoss, by default, collects garbage after 30 days, it would result in code that is much cleaner and thus more prone to bugs, which IMHO is more important value; any performance increase would be a nice bonus.

jtojnar:
Apparently the Firefox behaviour is correct one, since [Location.href](https://html.spec.whatwg.org/multipage/browsers.html#dom-location-hash) ought to return [URL’s fragment](https://url.spec.whatwg.org/#concept-url-fragment), which is described as an ASCII string.

Interestingly enough, [until 2015](https://www.fxsitecompat.com/en-CA/docs/2015/urlutils-hash-no-longer-decodes-fragment/) Firefox behaved the same way as Chromium. On the other hand, Safari, IE and old Opera are [correctly returning encoded hash](https://github.com/emberjs/ember.js/issues/3263), so this should probably be reported as Blink interop bug.

Since the `location.href` should be encoded everywhere, the proper fix is obtaining the hash using `decodeURIComponent(location.href.split('#').splice(1).join('#'))` ([source](https://stackoverflow.com/questions/1703552/encoding-of-window-location-hash#comment32913568_1704842)). It should be correct as, if I am reading the [RFC](https://tools.ietf.org/html/rfc3986#appendix-A) correctly, URL can only contain one number sign.

Decoding the hash as in d02a15916b4e22e4b5d0b3331f6dde42fb58d990 would fail in non-compliant browsers like chromium when the tag contains a percent sign.

niol:
Regarding _seek_ pagination, I do not understand why ```id``` should be included in the ```ORDER``` clause and in the comparison. Regarding performance, it seems about 5-10% better.

bendem:
Because multiple rows could have the same timestamp, you only want the ones after the last one you got.

jtojnar:
Indeed.

> Paging requires a deterministic sort order.

See a more detailed writeup about the *seek* pagination:

http://use-the-index-luke.com/sql/partial-results/fetch-next-page

public/js/selfoss-events-navigation.js
```diff
@@ -71,7 +71,7 @@ selfoss.events.navigation = function() {
         $(this).addClass('active');
         
         if($(this).hasClass('nav-tags-all')==false) {
-            location.hash = '#' + selfoss.filter.type + '/tag-' + $(this).find('span').html();
+            location.hash = '#' + selfoss.filter.type + '/tag-' + $(this).find('span').html().replace('%', '%25');
```
jtojnar: This should probably be `encodeURIComponent($(this).find('span').html())`. Works both in Chromium and Firefox.

niol: Yes, but I had preferred the browsers to show a unencoded string in the address bar. I've also tested with both browers.

jtojnar: I am afraid that is Chromium issue, Firefox displays the hash decoded. It would be nice if someone with access to Safari could test how it behaves there.

niol: It also works with the firefox I have (```Mozilla/5.0 (X11; Linux x86_64; rv:50.0) Gecko/20100101 Firefox/50.0```).

jtojnar:
In a brief inspection it looks really nice. However, 2999994cee01c51377168d12645c0ec1ab1679bf broke clicking the entry source to get to source listing. Two clicks are now needed.

jtojnar:
Looks nice.

And while I am nitpicking, when you click the entry source, the source in the sidebar will be highlighted [only after the listing finished loading](https://gfycat.com/ThinTenderHarvestmouse) – the other way around than when [clicking the source in the sidebar](https://gfycat.com/DifficultAdorableDowitcher).

niol:
Yes, but the only way to fix this would be to make two AJAX requests: the first to load the source navigation sidebar items, the second to load the new filtered list of items. The current implementation loads both ressources using the same request, which should be a bit faster.

SSilence:
This changes doesn't work with my sqlite database:

Ein Fehler ist aufgetreten: PDO: no such column: TRUE [selfoss/libs/f3/base.php:2032] Base->error(500,'PDO: no such column: TRUE') [selfoss/daos/mysql/Items.php:286] DB\SQL->exec('SELECT items.id FROM items AS items, sources AS sources WHERE items.source=sources.id AND TRUE LIMIT 1 OFFSET 50',array()) [selfoss/daos/Items.php:83] daos\mysql\Items->get(array('starred'=>false,'offset'=>'0','search'=>'','items'=>50,'offset_from_datetime'=>'','offset_from_id'=>'','itemsPerPage'=>'50','type'=>'newest','tag'=>'','source'=>'','sourcesNav'=>'false','ajax'=>'true')) [selfoss/controllers/Index.php:243] daos\Items->get(array('offset'=>'0','offset_from_datetime'=>'','offset_from_id'=>'','itemsPerPage'=>'50','search'=>'','type'=>'newest','tag'=>'','source'=>'','sourcesNav'=>'false','ajax'=>'true'))

PR #837, opened 2017-01-03
Open item link failed on mobile sometimes

Gugiman:
Couldn't find the exakt reason but sometimes I wasn't able to open item links with the "open" Button on mobile. The window opened with "about:blank" on Chrome (Android 5.1.1).

This is an example on how it works to me now.
Maybe someone knows there might be a problem on mobile devices and knows a better solution.

### DIFF
--- a/public/js/selfoss-events-entriestoolbar.js
+++ b/public/js/selfoss-events-entriestoolbar.js
@@ -21,7 +21,8 @@ selfoss.events.entriesToolbar = function(parent) {
     
     // open in new window
     parent.find('.entry-newwindow').unbind('click').click(function(e) {
-        window.open($(this).parents(".entry").children(".entry-datetime").attr("href"));
+        var itemLink = $(this).parents(".entry").children(".entry-datetime").attr("href");
+        window.open(itemLink);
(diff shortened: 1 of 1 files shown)


jtojnar:
This should really make no difference.

Gugiman:
I absolutely agree. I also added an alert() to see if there is just a problem fetching the url. After inserting the alert shows the correkt url **AND** the new tab opens the correct url.

My testcase:
I added this feed http://kubuntu.org/feed/ as a simple "RSS Feed".
Actually there's just one feed appearing. But I'm not able to open it. The new tab just calls "about:blank"

I'm not that into Javascript or jQuery so I didn't figured out if this is just a browser issue.

jtojnar:
It sounds like some weird race condition. Could you try to dump the following:

```
console.debug($(this));
console.debug($(this).parents(".entry"));
console.debug($(this).parents(".entry").children(".entry-datetime"));
console.debug($(this).parents(".entry").children(".entry-datetime").attr("href"));
```

https://developers.google.com/web/tools/chrome-devtools/remote-debugging/

PR #838, opened 2017-01-06
Automatically fetch feed title (Fix #662)

### LINKED ISSUE #662: Automatically fetch feed title
A lot of RSS readers that I've used will automatically pull in the title once the URL is pasted. I think this would be a great addition to selfoss.

niol:
A modest attempt to automatically fill-in source title.

### DIFF
--- a/controllers/Sources.php
+++ b/controllers/Sources.php
@@ -132,19 +132,25 @@ class Sources extends BaseController {
         // read data
         parse_str(\F3::get('BODY'),$data);
 
-        if(!isset($data['title']))
+        $autotitle = ($data['autotitle']=="1");
+
+        if(!$autotitle && !isset($data['title']))
             $this->view->jsonError(array('title' => 'no data for title given'));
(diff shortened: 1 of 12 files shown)


jtojnar:
I do not like this approach very much. IMHO, the following would be much cleaner:

1. Create temporary `Source` object from the `$params`.
2. If `$source->title` is empty:
    2.1. Call `$contentLoader->fetchTitle($source)`.
    2.2. `if ($title === null) $this->view->error('Missing title and cannot fetch one.');`
    2.3. `$source->setTitle($title)`
3. Persist the source.

Well, instead of objects, our DAOs return arrays for some reason; the rest stands, though.

SSilence:
I think a own parameter is a bit confusing for the user. The easier way would be following:
1. User creates a new source
2. User inserts the RSS URL
3. A ajax call will request from backend the title. The backend will try to fetch this title from base url given by rss feed url.
4. The result will be prefilled and the user can change this.

PR #844, opened 2017-01-14
Use bcrypt for password hashing

jtojnar:


bendem:
This is a good change imo, but it might be wort relying on `password_hash` instead if available.

jtojnar:
And now I noticed it has been marked as deprecated. I will switch to https://github.com/ircmaxell/password_compat once #845 is merged.

jtojnar:
#1137 will switch to `password_hash` for new passwords, while retaining the support for the legacy scheme.

PR #845, opened 2017-01-14
Composer

jtojnar:
Another go at #515. This time I kept the libraries not available on packagist (floIcon and content-extractor) in the repository – they are dead anyway.


SSilence:
Very cool, thank you a lot!

PR #846, opened 2017-01-15
Sync client from server using last update timestamp

niol:
This PR includes #834 .

Those changes add a new /items/sync API endpoint that sends database changes from a given datetime. This ensures only data absent from the client side is sent. This PR includes some client side changes to take advantage of this, including some functional enhancements such as cross-device updating of item statuses.

### DIFF
--- a/controllers/Items.php
+++ b/controllers/Items.php
@@ -169,4 +169,48 @@ class Items extends BaseController {
 
         $this->view->jsonSuccess($stats);
     }
+
+
+    /**
+     * returns updated database info (stats, item statuses)
+     * json
+     *
+     * @return void
+     */
+    public function sync() {
+        $this->needsLoggedInOrPublicMode();
+
+        if( !array_key_exists('since', $_GET) )
+            $this->view->jsonError(array('sync' => 'missing since argument'));
+
+        $since = new \DateTime($_GET['since']);
+
+        $itemsDao = new \daos\Items();
+        $last_update = new \DateTime($itemsDao->lastUpdate());
+
+        $sync = array(
+            'last_update' => $last_update->format(\DateTime::ISO8601),
+        );
+
+        if( $last_update > $since ) {
+            $sync['stats'] = $itemsDao->stats();
+
+            if( array_key_exists('tags', $_GET) && $_GET['tags'] == 'true' ) {
+                $tagsDao = new \daos\Tags();
+                $tagsController = new \controllers\Tags();
+                $sync['tagshtml'] = $tagsController->renderTags($tagsDao->getWithUnread());
+            }
+            if( array_key_exists('sources', $_GET) && $_GET['sources'] == 'true' ) {
+                $sourcesDao = new \daos\Sources();
+                $sourcesController = new \controllers\Sources();
+                $sync['sourceshtml'] = $sourcesController->renderSources($sourcesDao->getWithUnread());
+            }
+
+            $wantItemsStatuses = array_key_exists('items_statuses', $_GET) && $_GET['items_statuses'] == 'true';
+            if( $wantItemsStatuses ) {
+                $sync['items'] = $itemsDao->statuses($since->format(\DateTime::ISO8601));
+            }
+        }
+        $this->view->jsonSuccess($sync);
+    }
 }
--- a/controllers/Items.php
+++ b/controllers/Items.php
@@ -189,7 +189,7 @@ class Items extends BaseController {
         $last_update = new \DateTime($itemsDao->lastUpdate());
 
         $sync = array(
-            'last_update' => $last_update->format(\DateTime::ISO8601),
+            'last_update' => $last_update->format(\DateTime::ATOM),
         );
 
         if( $last_update > $since ) {
@@ -208,7 +208,7 @@ class Items extends BaseController {
 
             $wantItemsStatuses = array_key_exists('items_statuses', $_GET) && $_GET['items_statuses'] == 'true';
             if( $wantItemsStatuses ) {
-                $sync['items'] = $itemsDao->statuses($since->format(\DateTime::ISO8601));
+                $sync['items'] = $itemsDao->statuses($since->format(\DateTime::ATOM));
             }
         }
         $this->view->jsonSuccess($sync);
--- a/daos/mysql/Database.php
+++ b/daos/mysql/Database.php
@@ -229,4 +229,37 @@ class Database {
     public function optimize() {
         @\F3::get('db')->exec('OPTIMIZE TABLE `'.\F3::get('db_prefix').'sources`, `'.\F3::get('db_prefix').'items`');
     }
+
+
+    /**
+     * Ensure row values have the appropriate PHP type. This assumes we are
+     * using buffered queries (sql results are in PHP memory).
+     *
+     * @param expectedRowTypes associative array mapping columns to PDO types
+     * @param rows array of associative array representing row results
+     * @return array of associative array representing row results having
+     *         expected types
+     */
+    public function ensureRowTypes($expectedRowTypes, &$rows) {
+        foreach($rows as $rowIndex => $row) {
+            foreach($expectedRowTypes as $column => $type) {
+                if( array_key_exists($column, $row) ) {
+                    switch($type) {
+                        case \PDO::PARAM_INT:
+                            $value = intval($row[$column]);
+                            break;
+                        case \PDO::PARAM_BOOL:
+                            if( $row[$column] == "1" )
+                                $value = true;
+                            else
+                                $value = false;
+                            break;
+                    }
+                    // $row is only a reference, so we change $rows[$rowIndex]
+                    $rows[$rowIndex][$column] = $value;
+                }
+            }
+        }
+        return $rows;
+    }
 }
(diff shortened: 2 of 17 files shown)


jtojnar:
Nicer diff: https://github.com/niol/selfoss/compare/ppr/navigable...niol:ppr/sync

controllers/Items.php
```diff
@@ -170,4 +170,48 @@ public function stats() {
...
+        $last_update = new \DateTime($itemsDao->lastUpdate());
+
+        $sync = array(
+            'last_update' => $last_update->format(\DateTime::ISO8601),
+        );
+
+        if( $last_update > $since ) {
+            $sync['stats'] = $itemsDao->stats();
+
+            if( array_key_exists('tags', $_GET) && $_GET['tags'] == 'true' ) {
```
jtojnar: Now this is almost like GraphQL. Would not it be better to create a GraphQL endpoint? It could also save a lot requests during initial connection.

public/js/selfoss-events-entriestoolbar.js
```diff
@@ -104,7 +93,7 @@ selfoss.events.entriesToolbar = function(parent) {
                 type: 'POST',
                 error: function(jqXHR, textStatus, errorThrown) {
                     // rollback ui changes
-                    setButton(!starr);
+                    selfoss.ui.entryStarr(id, starr);
```
jtojnar: This should probably be `!starr`

public/js/selfoss-events-entries.js
```diff
@@ -151,8 +151,9 @@ selfoss.events.entries = function(e) {
             dataType: 'json',
```
jtojnar: After loads the last page of the feed, this request is fired on every scroll event. Probably because now the `stream-more` button is always present: https://github.com/niol/selfoss/blob/2620ee0c2a17ae44e4fffb63968f1cd5a23b3312/public/js/selfoss-events-entries.js#L130

public/js/selfoss-ui.js
```diff
@@ -46,6 +46,25 @@ selfoss.ui = {
                 selfoss.ui.entryMark(id, new_status.unread);
             }
         });
+    },
+
+
+    refreshStreamButtons: function(entries=false,
```
jtojnar: Nice, I did not know that any browsers already support this syntax. [IE does not](https://kangax.github.io/compat-table/es6/#test-default_function_parameters), though so we’d better stay at ES5 for now.

public/js/selfoss-ui.js
```diff
@@ -0,0 +1,71 @@
...
+            });
+            if( new_status ) {
+                selfoss.ui.entryStarr(id, new_status.starr);
+                selfoss.ui.entryMark(id, new_status.unread);
+            }
+        });
+    },
+
+
+    refreshStreamButtons: function(entries=false,
```
jtojnar: Nice, I did not know that any browsers already support this syntax. [IE does not](https://kangax.github.io/compat-table/es6/#test-default_function_parameters), though so we’d better stay at ES5 for now.

helpers/ViewHelper.php
```diff
@@ -70,6 +70,19 @@ public function dateago($datestr) {
...
+
+    /**
+     * Return ISO8601 formatted date
+     *
+     * @param sql date
+     * @return string
+     */
+    public static function date_iso8601($datestr) {
+        $date = new \DateTime($datestr);
+        return $date->format(\DateTime::ISO8601);
```
jtojnar: `\DateTime::ISO8601` is not actually ISO 8601 compliant, `\DateTime::ATOM` should be used: https://secure.php.net/manual/en/class.datetime.php#datetime.constants.atom

public/js/selfoss-ui.js
```diff
@@ -35,22 +35,28 @@ selfoss.ui = {
...
-    refreshItemStatuses: function(entry_statuses) {
+    refreshItemStatuses: function(entryStatuses) {
         $('.entry').each(function(index, item) {
             var id = $(this).data(('entry-id'));
-            new_status = entry_statuses.find(function(entry_status) {
-                return entry_status.id == id;
+            var newStatus = false;
+            entryStatuses.some(function(entryStatus) {
+                if( entryStatus.id == id )
+                    newStatus = entryStatus.id;
```
jtojnar: Should not this be just `newStatus = entryStatus`? And since you are actually trying to **find** some `entryStatus`, why do you use `Array.prototype.some`?

Spy-Seth:
Seeing this kind of PR, make me dream of a PWA front one day 👍

daos/mysql/Items.php
```diff
@@ -478,4 +478,32 @@ public function stats() {
...
+
+
+    /**
+     * returns the statuses of items last update
+     *
+     * @param date since to return item statuses
+     * @return array of unread, starred, etc. status of specified items
+     */
+    public function statuses($since) {
+        $res = \F3::get('db')->exec('SELECT id, unread, starred
```
jtojnar: Is it possible to use proper booleans? Otherwise the sync will mark all changed items as starred and unread since unlike in PHP `!!"0" === true` in JS:

```shell
$ php -r 'var_dump(!!"0");'
Command line code:1:
bool(false)

$ node -e 'console.log(!!"0")'
true
```

public/js/selfoss-ui.js
```diff
@@ -0,0 +1,77 @@
...
+            button.removeClass('active');
+            button.html($('#lang').data('unmark'));
+            parent.removeClass('unread');
+        }
+    },
+
+
+    refreshItemStatuses: function(entryStatuses) {
+        $('.entry').each(function(index, item) {
+            var id = $(this).data(('entry-id'));
```
jtojnar: Superfluous parentheses.

jtojnar:
1. Open a stream with only one page available.
2. Mark the first item as read.
3. In a different tab, mark the item as unread.
4. Wait for synchronization, “stream-more’ button will appear.
5. Click the button, nothing will be added since the item is above the offset.

niol:
85196bd1f55646109b3380bd90c92fbb9411efd0 should fix both of the reported issues that occur when using mysql or sqlite:
- the bool type in json is required for item statuses to propagate
- the int type for stat counts is required for the stream-more button not to show in the case described above

daos/mysql/Database.php
```diff
@@ -229,4 +229,37 @@ public function __construct() {
...
+                            else
+                                $value = false;
+                            break;
+                    }
+                    // $row is only a reference, so we change $rows[$rowIndex]
+                    $rows[$rowIndex][$column] = $value;
+                }
+            }
+        }
+        return $rows;
```
jtojnar: Why do you return the value though it is mutated in-place and the returned value is never used?

niol: I thought it might be more handy sometimes to directly return the call to `ensureRowTypes()`.

jtojnar: To be fair, I would prefer not using reference at all, unless the performance benefits are significant. Ideally, PHP engine would be smart enough to optimize it.

niol: While working on this, I though about this being used on big SQL results and I red about PHP arrays passed by value being duplicated in memory by the PHP engine when modified within the function. I wanted to prevent this. It is trivial to remove though.

jtojnar: For now, I would remove the returns, if we need them in the future, they can always be re-added. Though, in the long term I would like to drop this altogether and use some abstraction like https://nextras.org/orm/

jtojnar:
Sometimes when I use “refresh sources” button, after the refresh finishes, the view is not updated. I cannot reproduce it consistently, though, maybe there is some race condition with sync?

niol:
What do you mean by "the view is not updated"?

jtojnar:
I mean that the new items are not added, unless I reload the page.

niol:
For now, it only reloads if the list is empty. It only update stats and shows the read more button if applicable.

public/js/selfoss-base.js
```diff
@@ -228,36 +239,59 @@ var selfoss = {
...
+
+                var dataDate = new Date(data.last_update);
+
+                if( dataDate <= selfoss.lastUpdate )
+                    return;
+
+                if( data.stats.unread>0 &&
                     ($('.stream-empty').is(':visible') ||
                      $('.stream-error').is(':visible')) ) {
                     selfoss.reloadList();
```
jtojnar: I see now.

public/js/selfoss-events-navigation.js
```diff
@@ -173,12 +170,13 @@ selfoss.events.navigation = function() {
                     $('#nav-mobile-settings').click();
                     
                 // refresh list
-                 selfoss.reloadList();
+                selfoss.sync(true);
```
jtojnar: It should be fixed so the user does not have to reload after the refresh.

niol: Refreshing sources could take some time. The point of the change is to let that run in the background and let the user continue reading and reload list ONLY if it is empty. If list has to be reloaded when refreshing is done even if items may still be displayed thus red by the user, I would rather revert 3071660c7c600aa68562f82c9e9db36ed7ca7867 .

jtojnar: Good point. I got used to waiting for the page to reload from previous versions but this might be actually better. In the future, it would be nice to add UI for reloading the list.

jtojnar:
It looks good to me. Any points before merging this?

PR #849, opened 2017-01-22
Add JSON support for Sources::write

### LINKED ISSUE #734: [Question] How to pass the POST parameters to the API?
Hi,
I have one question that is barely documented in the Wiki. **In which format do I need to submit the parameters?**

For example, I'm trying to add a subscription, so I perform a POST request to `../source`. All the required fields are submitted as JSON object like the following example.

``` json
{
    "title": "Neues Deutschland",
    "spout": "spouts\\rss\\feed",
    "tags": "",
    "params": {
        "url": "http://neues-deutschland.de/rss/neues-deutschland.xml"
    }
}
```

However, selfoss only returns HTTP code 400 with the message `{"title":"no data for title given"}`. **And how can I transmit the credentials of the user in the example above, if there are any?**

jtojnar:
Fixes #734 for `/source` and `/source/@id` endpoints. Requires f3 3.6.0 to work (use #845).

### DIFF
--- a/controllers/Sources.php
+++ b/controllers/Sources.php
@@ -130,7 +130,13 @@ class Sources extends BaseController {
         $sourcesDao = new \daos\Sources();
 
         // read data
-        parse_str(\F3::get('BODY'),$data);
+        $headers = \F3::get('HEADERS');
+        $body = \F3::get('BODY');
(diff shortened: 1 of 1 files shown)


jtojnar:
@SSilence note that this requires newer F3. Merging composer first would be better.

PR #851, opened 2017-01-29
Automatically fetch feed title

jtojnar:
This is my attempt at automatic fetching of title.

The title is fetched when saving the spout without title filled in. Unlike #838, this also has complete spout coverage.

cc  @niol

### DIFF
--- a/controllers/Sources.php
+++ b/controllers/Sources.php
@@ -132,8 +132,19 @@ class Sources extends BaseController {
         // read data
         parse_str(\F3::get('BODY'),$data);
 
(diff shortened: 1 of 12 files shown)


SSilence:
Thanks for the pr. I have a bug: after saving no title will be shown. I must refresh the source list for seeing the fetched title.

jtojnar:
@SSilence Do you have debug mode disabled or did you not remove the `all.js` file?

SSilence:
Ah you are right, the browser cache. It works fine, thanks a lot!

PR #869, opened 2017-02-06
Make selfoss navigable

niol:
These patches make selfoss navigable using the url hash. This makes the url reflect whats displayed in selfoss. This also makes it possible to reload the item that was open before a browser crash or when the tab containing selfoss was offloaded from memory on a mobile browser.

This is #834 rebased and with two additional fixes.

### DIFF
--- a/daos/mysql/Items.php
+++ b/daos/mysql/Items.php
@@ -239,8 +239,12 @@ class Items extends Database {
             && isset($options['offset_from_id'])
             && is_numeric($options['offset_from_id']) ) {
 
+            $offset_from_datetime_sql = $this->stmt->datetime($options['offset_from_datetime']);
             $params[':offset_from_datetime'] = array(
-                $options['offset_from_datetime'], \PDO::PARAM_STR
+                $offset_from_datetime_sql, \PDO::PARAM_STR
+            );
+            $params[':offset_from_datetime2'] = array(
+                $offset_from_datetime_sql, \PDO::PARAM_STR
             );
             $params[':offset_from_id'] = array(
                 $options['offset_from_id'], \PDO::PARAM_INT
@@ -251,7 +255,12 @@ class Items extends Database {
             else
                 $ltgt = '<';
 
-            $where[] = "(items.datetime, items.id) $ltgt (:offset_from_datetime, :offset_from_id)";
+            // Because of sqlite lack of tuple comparison support, we use a
+            // more complicated condition.
+            $where[] = "(items.datetime $ltgt :offset_from_datetime OR
+                         (items.datetime = :offset_from_datetime2 AND
+                          items.id $ltgt :offset_from_id)
+                        )";
         }
 
         $where_ids = '';
--- a/daos/mysql/Items.php
+++ b/daos/mysql/Items.php
@@ -193,7 +193,7 @@ class Items extends Database {
      */
     public function get($options = array()) {
         $params = array();
-        $where = array('TRUE');
+        $where = array($this->stmt->bool(true));
         $order = 'DESC';
                 
         // only starred
--- a/daos/mysql/Items.php
+++ b/daos/mysql/Items.php
@@ -306,6 +306,14 @@ class Items extends Database {
         $order = 'ORDER BY items.datetime,items.id '.$order;
 
         if( $where_ids != '' ) {
+            // This UNION is required for the extra explicitely requested items
+            // to be included whether or not they would have been excluded by
+            // seek, filter, offset rules.
+            //
+            // SQLite note: the 'entries' SELECT is encapsulated into a
+            // SELECT * FROM (...) to fool the SQLite engine into not
+            // complaining about 'order by clause should come after union not
+            // before'.
             $query = "SELECT * FROM (
                         SELECT * FROM ($select $where_sql $order LIMIT " . $options['items'] . ' OFFSET '. $options['offset'] . ") AS entries
                       UNION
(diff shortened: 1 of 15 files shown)


jtojnar:
Another problem:

1. Open selfoss
2. Click settings

Selfoss will change URL to http://localhost/selfoss/#sources/all

Which reminds me that there is no unknown route handling – if I go to http://localhost/selfoss/#newest/source-99999, it will just show empty listing. Though the API is partly to blame because it does not return an error on invalid source filter.

jtojnar:
Okay, this fixes clicking on settings and then filters. Settings followed by tags are still broken.

jtojnar:
Much better. Also, when clicking settings and then filter, no tag or source will be selected.

jtojnar:
Looks good so far. Could you also handle non-existent sources and tags and invalid links?

* http://localhost/selfoss/#unread/source-99999
* http://localhost/selfoss/#sourcen

jtojnar:
I am now testing it on my live server and noticed another detail:

1. Select “All tags”
2. Click any other tag or source.
3. Use browser “Back” function

The “All tags” filter will not be highlighted in the sidebar. Tags or sources do not suffer from this. The navigation works, the issue is purely visual.

jtojnar:
Also it seems like SQLite does not support tuple comparisons. You will need to expand it as follows:

```sql
(items.datetime $ltgt :offset_from_datetime OR (items.datetime = :offset_from_datetime AND items.id $ltgt :offset_from_id))
```

jtojnar:
Nice. Is it intentional that the URL is only changed when an item is clicked, not when keyboard shortcut is used?

jtojnar:
Neat, though you still cannot navigate back/forward for these.

jtojnar:
Additionally, is the extra_ids thing still working? When I open an item, for example, on the second page, then scroll up and reload, it is not visible on the first page. Nevertheless, the item comes up as a first item on every page after its own page.

jtojnar:
Isn’t the item ID in the URL completely useless now?

jtojnar:
Thanks, now only remove the `extra_ids` code.

* when the id of the item is on the first page, it will be already matched by the first part of `$where_sql`
* when it is on any other page, it will be matched by the `$where_ids` part but discarded by the limit.

niol:
For me, as said, the whole point of putting the item id in the url is that it makes it possible to reload the item that was open before a browser crash or when the tab containing selfoss was offloaded from memory. Without this, the problem was that the item was lost in read items and that it was a lot of searching to find it back, for instance to star it or to finish reading.
The ```extra_ids``` parameter is a key part in making this work.

jtojnar:
The problem is that `exra_ids` only works when `$where_sql` is `false`; that is when the item is read, you are viewing unread items and there is less than maxPageLimit of them.

jtojnar:
In other cases, because the items are sorted by `datetime, id`, the `extra_ids` items will be pushed to next pages by items with higher key.

daos/mysql/Items.php
```diff
@@ -247,16 +291,16 @@ public function get($options = array()) {
...
 
         // get items from database
-        return \F3::get('db')->exec('SELECT 
+        return \F3::get('db')->exec('SELECT
                     items.id, datetime, items.title AS title, content, unread, starred, source, thumbnail, icon, uid, link, updatetime, author, sources.title as sourcetitle, sources.tags as tags
                    FROM '.\F3::get('db_prefix').'items AS items, '.\F3::get('db_prefix').'sources AS sources
-                   WHERE items.source=sources.id '.$where.' 
-                   ORDER BY items.datetime '.$order.' 
+                   WHERE items.source=sources.id AND '.$where_sql.'
+                   ORDER BY items.datetime '.$order.'
```
jtojnar: Actually they are only ordered by `datetime`, this should be also ordered by `id` for seek pagination to work properly.

niol:
You're right, I'll see if I can fix that using only one query.

daos/mysql/Items.php
```diff
@@ -296,12 +299,32 @@ public function get($options = array()) {
...
-                    items.id, datetime, items.title AS title, content, unread, starred, source, thumbnail, icon, uid, link, updatetime, author, sources.title as sourcetitle, sources.tags as tags
-                   FROM '.\F3::get('db_prefix').'items AS items, '.\F3::get('db_prefix').'sources AS sources
-                   WHERE items.source=sources.id AND '.$where_sql.'
-                   ORDER BY items.datetime '.$order.'
-                   LIMIT ' . $options['items'] . ' OFFSET '. $options['offset'], $params);
+        $select = 'SELECT
+            items.id, datetime, items.title AS title, content, unread, starred, source, thumbnail, icon, uid, link, updatetime, author, sources.title as sourcetitle, sources.tags as tags
+            FROM '.\F3::get('db_prefix').'items AS items, '.\F3::get('db_prefix').'sources AS sources
+            WHERE items.source=sources.id AND';
+        $order = 'ORDER BY items.datetime,items.id '.$order;
```
jtojnar: This should probably be `'ORDER BY items.datetime ' . $order . ', items.id ' . $order`

jtojnar:
If I open page at other than default filter (e.g. https://localhost/selfoss/#starred/all), `selfoss.filter.type` will be `unread`.

jtojnar:
This looks good, if you are okay with that, I will rebase it and merge.

niol:
Many thanks for your testing.
Merging should not need rebasing.

PR #870, opened 2017-02-06
Coding style

jtojnar:
This will be merged once all other pending PRs are merged.


controllers/Index.php
```diff
@@ -228,50 +227,53 @@ private function loadItems($options, $tags) {
...
         }
 
-        if(strlen($itemsHtml)==0) {
-            $itemsHtml = '<div class="stream-empty">'. \F3::get('lang_no_entries').'</div>';
+        if (strlen($itemsHtml) == 0) {
+            $itemsHtml = '<div class="stream-empty">' . \F3::get('lang_no_entries') . '</div>';
         } else {
-            if($itemDao->hasMore())
-                $itemsHtml .= '<div class="stream-more"><span>'. \F3::get('lang_more').'</span></div>';
-                $itemsHtml .= '<div class="mark-these-read"><span>'. \F3::get('lang_markread').'</span></div>';
```
jtojnar: Judging by the indentation, this was meant to be part of the `if`. Am I right, @niol?

Original commit: https://github.com/SSilence/selfoss/commit/3c1dc1965f6430a69584fbf75c1148f843d11749

niol: No indention is wrong here: the button should be shown even if there are no more items to show, because it is intended to mark visible items as read.

Note: this code is removed by #846.

jtojnar: Okay, never mind then.

PR #872, opened 2017-02-08
Add syntax linting using Travis

jtojnar:
We are linting the code on all supported versions of PHP.

Note that we remove TwitterOAuth library when running PHP older than 5.6 because TwitterOAuth declares 5.6 as minimal version. Except for one small detail, I have filed a pull request for, it still works, though.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,32 @@
+language: php
+
+matrix:
+    include:
+        - php: 7.1
+        - php: 7.0
+        - php: 5.6
+        - php: 5.5
+          env: REMOVE_TWITEROAUTH=true
+        - php: 5.4
+          env: REMOVE_TWITEROAUTH=true
+    fast_finish: true
+
+cache:
+    directories:
+        - vendor
+        - $HOME/.composer/cache
+
+before_install:
+    - if [ -n "$GH_TOKEN" ]; then composer config github-oauth.github.com ${GH_TOKEN}; fi
+    - composer create-project --no-dev --no-interaction jakub-onderka/php-parallel-lint utils/php-parallel-lint
+
+install:
+    - composer self-update
+
+before_script:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -0,0 +1,32 @@
+language: php
+
+matrix:
+    include:
+        - php: 7.1
+        - php: 7.0
+        - php: 5.6
+        - php: 5.5
+          env: REMOVE_TWITEROAUTH=true
+        - php: 5.4
```
Spy-Seth: PHP 5.4 is not supported anymore: http://php.net/supported-versions.php

jtojnar: Selfoss still supports PHP 5.4. As does [Debian Wheezy][1], which is on LTS until [May 2018][2].

[1]: https://wiki.debian.org/PHP#Available_versions
[2]: https://wiki.debian.org/LTS

Spy-Seth: Having Debian 7 in active LTS does not means PHP is still maintained: security fix are not back ported. Using PHP 5.4 on a server is not secure anymore (even on Debian 7).

Dropping the PHP 5.4 and 5.5 support could only make less maintenance work for you guys. People that need the PHP 5.4 (or 5.5) support could still use the older release.

jtojnar: The security fixes [ARE backported](https://wiki.debian.org/PHP#Notes_on_PHP_and_security). Apparently many shared hostings are running on Wheezy so there is at least some commercial interest in security of PHP 5.4.

Trust me, if I could I would drop PHP 5 altogether, it is not my call though.

Spy-Seth: I'm not convince, old release are still here... but that's your call ;)

jtojnar: Well, more like Tobias’s, I am just a co-maintainer.

Issue #858, opened 2017-02-04
Spout transformers


jtojnar:
Currently there are some derived spouts like “RSS (with Readability)“ or “RSS (with FullTextRss)“. It would be nice if there were spout transformers that could modify the output of the spouts. Then it would be possible to fix #705 simply by applying FullTextRss transformer to Reddit spout.

Additionally, instead of just `map`, the transformers could implement `filterMap` thus allowing smarter filters for spouts.

lyda:
Would this relate to #606?

jtojnar:
@lyda Yes. The current filters only use a single regex. Spout transformers would allow setting multiple filters. Something like this:

![screenshot-2017-9-15 selfoss 2637](https://user-images.githubusercontent.com/705123/30482110-14f388d8-9a22-11e7-9e23-b0489d55f31e.png)

jtojnar:
I initially wanted to add advanced filters as part of this but since it is a feature many people including me want, I decided to split it out into #1125.

PR #885, opened 2017-03-05
Add ZDF Mediathek spout

PhrozenByte:
The RSS feed of the ZDF Mediathek (the VOD service of a german TV channel) often lists videos which actually haven't been released yet. This spout ignores such items.

Example: https://www.zdf.de/rss/zdf/wissen/leschs-kosmos lists various videos for the next show airing on Tuesday.

### DIFF
--- a/spouts/rss/zdf.php
+++ b/spouts/rss/zdf.php
@@ -1,0 +1,50 @@
+<?PHP 
+
+namespace spouts\rss;
+
+/**
+ * Plugin for fetching media from the ZDF Mediathek
+ *
+ * @package    plugins
+ * @subpackage news
+ * @copyright  Copyright (c) Daniel Rudolf
+ * @license    GPLv3 (http://www.gnu.org/licenses/gpl-3.0.html)
+ * @author     Daniel Rudolf <http://daniel-rudolf.de/>
+ */
+class zdf extends feed
+{
+    /**
+     * name of spout
+     *
+     * @var string
+     */
+    public $name = 'ZDF Mediathek';
+
+
+    /**
+     * description of this source type
+     *
+     * @var string
+     */
+    public $description = 'This feed fetches media items from the ZDF Mediathek.';
+
+    /**
+     * loads the contents of the feed
+     *
+     * @param mixed $params the params of this source
+     * @return void
+     */
+    public function load($params)
+    {
+        parent::load($params);
+
+        foreach ($this->items as $key => $item) {
+            // remove future items
+            $time = $item->get_date('U');
+            if ($time && ($time > time())) {
+                unset($this->items[$key]);
+            }
+        }
+    }
+
+}


jtojnar:
I am not sure we should add any more of such specific spouts. Personally, I would even like to remove the existing ones like teltarif. Instead, I would like to simplify writing spouts (#858) and provide a stable API so that users could write their own spouts and maintain them outside the core source tree.

PhrozenByte:
Afaik Selfoss is no software supposed to be used by developers only, thus this IMO is no good idea, because it will lead to significant worsening for most users. Furthermore, developers that use Selfoss will re-invent the wheel over and over again...

However, you've a good point: Things like what this spout does is pretty generic, most spouts consist of just a single change compared to the main RSS spout. A list of "transformers", as you call them, might be a good idea, like toggling a button that says "remove future items". It should be possible to group a set of transformers, so users don't have to think about what transformers they need - they select the "RSS" spout, enter a URL to the ZDF Mediathek and Selfoss should automatically select the correct transformers (the "remove future items" transformer in this case). Listing "ZDF Mediathek" next to "RSS" is weird.

Nevertheless, until this *planned* feature is implemented, it shouldn't interfere with *current* development.

jtojnar:
Of course, I do not expect non-developer users to write their own spouts; I would, however, like to encourage developers like yourself to create spouts and add them to a list of third-party spouts (probably under https://github.com/SSilence/selfoss/wiki#modifications). Other users would then just copy the spout to the spouts directory (or even install it from software repositories like AUR). This is how it might look like: https://github.com/fossar/selfoss-spout-teltarif

If the spout works without any selfoss modifications, it should be possible today. Well, except for one small problem with auto-loading, I will need to fix it before 2.17 is released.

PhrozenByte:
I don't see any advantage in this approach, it rather makes things much harder. Users have to search for and manually "install" spouts, a process that is probably already to hard for most everyday users, i.e. Selfoss's target audience. It even makes things harder for developers: they will have to create and maintain a repo for - yeah, just a spout? That's far away from being appropriate for things like that. I can speak just for myself, but I definitely won't do that, I would rather not publish the spout instead. This is neither User- nor Contributor-friendly.

jtojnar:
The reason is we do not have the capacity to maintain every spout we receive as a PR. Not only this creates a burden for us, it is annoying for users, who have to choose from the list of spouts – though this could be mostly fixed by a searchable select.

End users: Installing the additional spouts would be done the same way as selfoss was installed – by copying files into a directory, or preferably using the distribution’s install command like `yaourt -S selfoss-spout-teltarif`. User should therefore be already familiar with the process.

Developers: I understand a repo might be a bit overkill for spouts like this but you can host the spout on [gist](https://gist.github.com/) as well. Just add a link to wiki and user will get it from there.

PhrozenByte:
> The reason is we do not have the capacity to maintain every spout we receive as a PR.

At the moment this is rather a theoretical problem and definitely no actual problem. I can remember just two spouts that had problems for *years*...

Again: I see your point. However, until we have a better solution, like the mentioned "transformers" feature that can solve this problem with "magic" like evaluating feed URLs, I strongly advise against changing anything about how spouts are managed.

> End users

Please keep in mind that the person who installed Selfoss isn't necessarily the same person as the one who wants to add a spout. Furthermore, please don't forget that many users don't have access to the command line. Again, we're talking about *everyday* users, not developers. Sure, this is easy for people like us, but definitely not for all people I recommended Selfoss to.

> Developers

It will result in a huge mess of duplicate, deprecated and unmaintained spouts... 😒

jtojnar:
> I can remember just two spouts that had problems for years

Ignoring Facebook and YouTube, the latter recently fixed, there were actually quite a few problems. To name some, broken redirects in reddit spout, cookie not being saved in reddit spout, readability dying, weird unicode hacks in some other spouts, the problems with twitter. Well, these are actually all the popular spouts. I do not know about bugs in the obscure spouts like teltarif because I do not use them (I reckon few people do)

> Please keep in mind that the person who installed Selfoss isn't necessarily the same person as the one who wants to add a spout.

I am unsure how prevalent the assisted installs are but this is a valid point. One way to solve this would be by creating an interface for searching and installing spouts within selfoss – something like what web browsers have. We need to keep the distinction between officially supported spouts and third party ones, though.

> Duplicate, deprecated and unmaintained spouts

Yes, this is a problem but it is already here. The only difference is that the unmaintained spouts are in the core repository. And now we cannot remove them because someone might be using them.

PhrozenByte:
> there were actually quite a few problems

The question is: Do you think that managing all spouts in external repos improves the situation for Selfoss *users*? All these problems don't disappear, they rather worsen. You just hide behind "it's a third-party spout" - what definitely isn't wrong, you're completely right, however, it doesn't help.

The result will be that spouts are managed in external repos maintained by, yeah nobody. Fixes from third-party developers won't get merged into said external repos, what results in a mess of forks with different development states - or even the 100. independent development of the same spout.

This confuses users, in the worst case they will have to try multiple spouts before finding a working one - and they will one-hundred percent *blame Selfoss*! This won't happen when all spouts are managed in the core repo, it gives third-party developers the opportunity to fix a broken spout, otherwise its usually just the original author.

> One way to solve this would be by creating an interface for searching and installing spouts within selfoss

That's a even more complex solution. I still go for the "transformers" idea with a simple (built-in) detection of some known URLs - this should greatly reduce the number of spouts and managing the remaining spouts in the core repo shouldn't be a big deal.

> Yes, this is a problem but it is already here. The only difference is that the unmaintained spouts are in the core repository. And now we cannot remove them because someone might be using them.

The difference is that third-party developers have a chance to fix it, what isn't the case otherwise. However, you can definitely remove broken spouts - when they are broken they don't work anyway. Simply add a "noop spout" that prevents further updates, but keeps the existing items untouched and change removed spouts to this "noop spout" on updates.

jtojnar:
> Do you think that managing all spouts in external repos improves the situation for Selfoss users?

I think the situation will be the same for most users.

> Fixes from third-party developers won't get merged into said external repos, what results in a mess of forks with different development states - or even the 100. independent development of the same spout.

While hardly specific to selfoss, I agree that this is a problem. It would work better if  GitHub supported taking over dead repositories. One solution would be creating a central repository (e.g. fossar/selfoss-extra-spouts) where every spout would have to have a maintainer assigned to be kept there.

> This confuses users, in the worst case they will have to try multiple spouts before finding a working one - and they will one-hundred percent blame Selfoss! This won't happen when all spouts are managed in the core repo, it gives third-party developers the opportunity to fix a broken spout, otherwise its usually just the original author.

This would also be mitigated by the central repository. Though the list of spouts on the wiki should definitely help with that problem.

> That's a even more complex solution. I still go for the "transformers" idea with a simple (built-in) detection of some known URLs - this should greatly reduce the number of spouts and managing the remaining spouts in the core repo shouldn't be a big deal.

Yes, it would be a long term solution for simplifying the spout installation; it does not rule out the transformers.

> However, you can definitely remove broken spouts - when they are broken they don't work anyway. Simply add a "noop spout" that prevents further updates, but keeps the existing items untouched and change removed spouts to this "noop spout" on updates.

Here again, people would blame selfoss. If they install third party spout it should not go on our head.

----

Actually, I think there could be many spouts that are useful to lot of people but not enough to warrant inclusion to the main repo. For example, I have an [ello](https://ello.co/) spout on my to-do list but I do not think it deserves to be merged.

The whole issue is about responsibility. I do not want it 😜 

----

Additionally, could you try mailing ZDF about the unreleased items?

PhrozenByte:
> One solution would be creating a central repository (e.g. fossar/selfoss-extra-spouts) where every spout would have to have a maintainer assigned to be kept there.

So, where's the difference to merging them into the core repo?

> Here again, people would blame selfoss. If they install third party spout it should not go on our head.

I think this is the point. Selfoss has a ecosystem and third-party spouts are part of this ecosystem. It makes absolutely no difference how you manage third-party spouts or how you name them, it still is Selfoss's ecosystem - and Selfoss is one-hundred percent responsible for its ecosystem. If something is wrong with a spout, users will *always blame Selfoss*, no matter the spout is shipped together with Selfoss or managed in a third-party repo found in Selfoss's wiki (no matter if it is a central repo or not). You won't change that by outsourcing spouts, it always was this way and will always be, this is the reason why all bigger open-source projects have a more or less sure eye on their ecosystem. Users simply don't know *why* something isn't working, that's the reason why they are asking - they just see that Selfoss isn't doing what they want it to do, thus they blame Selfoss.

I support your idea that Selfoss should keep the number of spouts small - and the "transformers" idea is a pretty good solution for that. Besides that I still strongly advise against changing anything.

> The whole issue is about responsibility. I do not want it 😜

Yeah, I completely understand that - from a pretty practical point of view, I also have this responsibility and it would be great if I could hand it over. Anyway, handing over responsibility to "nobody" doesn't work, that's not how "responsibility" works :wink: (it would be great if that were the case, life would be so easy :smile:)

jtojnar:
> So, where's the difference to merging them into the core repo?

There would be clear distinction between supported and unsupported spouts.

> It makes absolutely no difference how you manage third-party spouts or how you name them, it still is Selfoss's ecosystem - and Selfoss is one-hundred percent responsible for its ecosystem. If something is wrong with a spout, users will always blame Selfoss, no matter the spout is shipped together with Selfoss or managed in a third-party repo found in Selfoss's wiki (no matter if it is a central repo or not).

While I agree that we are, to a certain point, responsible for its selfoss’s ecosystem, I would place the responsibility somewhere around 60%. Maybe it’s a wishful thinking, but I hope our users realize we are people too and that we cannot handle everything.

> I support your idea that Selfoss should keep the number of spouts small - and the "transformers" idea is a pretty good solution for that. Besides that I still strongly advise against changing anything.

I would like to close this PR so that we do need to remove this later when transformers are implemented.

PR #887, opened 2017-03-08
Add a sharing option to use Wallabag version 2.X

jauninp:
Add link to share with wallabag V2.x, by default it take the configuration of wallabag2 if the field is fill.


defaults.ini
```diff
@@ -31,6 +31,7 @@ use_system_font=
 readability=
 share=gtfprde
 wallabag=
+wallabag2=
```
jtojnar: Since only one of the wallabag keys will be used, could you change `wallabag2` to `wallabag_version`?

jtojnar:
Also, could you add the new key to the documentation?

jtojnar:
Thank you, merged as 56d2829c24c601df9685060273499ae25b2fbdaa

jtojnar:
Hi @jauninp. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contribution under `GPL-3-or-later`?

Issue #879, opened 2017-02-17
Use proper types in API


jtojnar:
Booleans should be `true` or `false`, not `"0"` or `"1"`. Similarly, numbers should not be transmitted as strings. Tags should be returned as a list, not a string of comma separated names.

niol:
Some work about this is available at https://github.com/niol/selfoss/tree/ppr/apitypes . As it seems what is wanted here is a major rework of how selfoss is driven from JS (like not returning HTML but lists for instance), I will stop there for now and let new features I want to work on drive those changes (using new endpoints).

Maybe a spec should be worked on before or some common RSS reader API be used instead.

jtojnar:
I would not go as far as using JSON instead of HTML now. I meant the tags in `/items` endpoint used by applications.

In the future building the UI client-side from the same endpoints applications use could be possible – that is actually one of the points I investigate in álmafoss – I do not consider it to be in the scope of this issue, though.

I have sent you an invite to álmafoss private repo in case you are interested. It is still very rough draft and I will not have time to finish it until June.

niol:
I fixed the tags in ```/items``` in my branch.
Thanks for the invite, but for now, I will use my free time for new features I want rather than rewriting the frontend and learning a new development environment.

jtojnar:
Mostly fixed in https://github.com/SSilence/selfoss/pull/948

PR #889, opened 2017-03-14
Various fixes and enhancements

niol:
Details are in the individual patches.


public/js/selfoss-ui.js
```diff
@@ -71,6 +71,53 @@ selfoss.ui = {
...
+
+
+    /**
+     * show error
+     *
+     * @return void
+     * @param message string
+     */
+    showError: function(message) {
+        selfoss.ui.showMessage(message, 'undefined', 'undefined', true);
```
jtojnar: This should probably be quoteless, since `typeof 'undefined' === 'string'`, leading to showing button with undefined text in error messages (e.g. invalid fragment like `#invalid`)

jtojnar:
I am not sure if #879 should be fixed right now, since it is a BC break. I will ask Tobias about BC break policy.

jtojnar:
For now, it would be probably better to leave out the commit since we are essentially on 2.17 feature freeze.

jtojnar:
Also there are lot of API endpoints still affected, separate pull request will probably be for the best.

jtojnar:
Another issue with the async refresh: refreshing on the sources page will keep loading forever with the error `selfoss.lastUpdate is null` in the console.

jtojnar:
Another point against the API change: 4eb189198d5090261e1743d6494dccacab42db58 breaks adding sources.

controllers/Sources.php
```diff
@@ -165,7 +165,7 @@ public function write() {
         unset($data['ajax']);
 
         // check if source already exists
-        $id = \F3::get('PARAMS["id"]');
+        $id = intval(\F3::get('PARAMS["id"]'));
```
jtojnar: This will break the source adding since `intval('new-123') === 0`, which is a valid ID according to `$sourcesDao->isValid`.

daos/mysql/Sources.php
```diff
@@ -144,13 +144,15 @@ public function get($id = null) {
         // select source by id if specified or return all sources
         if (isset($id)) {
             $ret = \F3::get('db')->exec('SELECT id, title, tags, spout, params, filter, error FROM ' . \F3::get('db_prefix') . 'sources WHERE id=:id', [':id' => $id]);
+            $this->stmt->ensureRowTypes($ret, ['id' => \PDO::PARAM_INT]);
```
jtojnar: This should be assigned to `$ret`.

daos/mysql/Sources.php
```diff
@@ -144,13 +144,15 @@ public function get($id = null) {
...
             if (isset($ret[0])) {
                 $ret = $ret[0];
             } else {
                 $ret = false;
             }
         } else {
             $ret = \F3::get('db')->exec('SELECT id, title, tags, spout, params, filter, error FROM ' . \F3::get('db_prefix') . 'sources ORDER BY error DESC, lower(title) ASC');
+            $this->stmt->ensureRowTypes($ret, ['id' => \PDO::PARAM_INT]);
```
jtojnar: This should be assigned to `$ret`.

public/css/style.css
```diff
@@ -44,6 +44,23 @@ body * {
     padding:10px;
 }
 
+#message a {
+    font-size:0.9em;
+    text-align:center;
+    -moz-border-radius:2px;
```
jtojnar: No need for prefixed versions, since border-radius is [widely supported](http://caniuse.com/#feat=border-radius).

jtojnar:
Additionally, 00ced63f75d31503f0c9fdf8b7824ba84637464f broke error handling when adding source. Try adding http://example.com as an RSS feed with no title: the server will correctly return error but client will keep loading.

public/js/selfoss-base.js
```diff
@@ -261,7 +264,7 @@ var selfoss = {
                 since:          selfoss.lastUpdate.toISOString(),
                 tags:           true,
                 sources:        selfoss.filter.sourcesNav,
-                items_statuses: true
+                items_statuses: selfoss.lastUpdate !== null
```
jtojnar: Won’t this now be always true, since there is line 256?

public/css/style.css
jtojnar: The prefixed border-radius was removed everywhere but here.

jtojnar:
Merged as 6beba94934bf5b44c2666a9b2e10ade050503575

Issue #895, opened 2017-03-17
2.17 breaks full-text-rss with call to Readability


Alrescha-git:
After upgrading to 2.17, updates failed with message:

"selfoss.ERROR: fatal error: Class 'Readability' not found"

I do not use Readability for anything, but I did find a call to Readability() in the full-text RSS spout.  Switching those feeds to just RSS solved the update issue.

rwky:
Hah you beat me to it by 3 minutes, also reported in #896

jtojnar:
Thanks for reporting. Here is an [updated version](https://bintray.com/fossar/selfoss/selfoss-git/2.18-788d954#files).

Binnette:
I think I'm also affected. Here is my error log:
```
[2017-11-27 21:13:38] selfoss.INFO: ftrss - Loading page: http://www.developpez.com/actu/175734/Le-passage-automatique-en-mode-de-navigation-privee-bientot-une-realite-dans-les-navigateurs-de-Microsoft-La-firme-depose-une-demande-de-brevet/
[2017-11-27 21:13:38] selfoss.INFO: ftrss - Extracting content
[2017-11-27 21:13:38] selfoss.DEBUG: ftrss - Cleaning content
[2017-11-27 21:13:38] selfoss.DEBUG: ftrss - Could not find Content-Type header in HTTP response
[2017-11-27 21:13:38] selfoss.DEBUG: ftrss - Character encoding: iso-8859-1
[...]

PR #897, opened 2017-03-18
Use Graby instead Full-Text-Rss

jtojnar:
Closes: #895


Weishaupt:
Can you explain, why you removed the git submodule? Do I need to put every site I want to fetch in the custom directory?

jtojnar:
Graby uses composer to fetch the data that was previously obtained using a submodule. You can find it in `vendor/j0k3r/graby-site-config/`.

PR #906, opened 2017-03-26
Various fixes

niol:
Various fixes detailed in the individual commits.

### DIFF
--- a/spouts/rss/feed.php
+++ b/spouts/rss/feed.php
@@ -212,7 +212,7 @@ class feed extends \spouts\spout {
      */
     public function getTitle() {
         if ($this->items !== false && $this->valid()) {
-            return @current($this->items)->get_title();
+            return htmlspecialchars_decode(@current($this->items)->get_title());
         }
 
         return false;
@@ -300,9 +300,9 @@ class feed extends \spouts\spout {
             if (isset($author)) {
                 $name = $author->get_name();
                 if (isset($name)) {
-                    return $name;
+                    return htmlspecialchars_decode($name);
                 } else {
-                    return $author->get_email();
+                    return htmlspecialchars_decode($author->get_email());
                 }
             }
         }
--- a/spouts/spout.php
+++ b/spouts/spout.php
@@ -91,14 +91,18 @@ abstract class spout implements \Iterator {
     abstract public function getId();
 
     /**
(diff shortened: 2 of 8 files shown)


spouts/rss/feed.php
```diff
@@ -212,7 +212,7 @@ public function getId() {
      */
     public function getTitle() {
         if ($this->items !== false && $this->valid()) {
-            return @current($this->items)->get_title();
+            return htmlspecialchars_decode(@current($this->items)->get_title());
```
jtojnar: Why are you decoding the title?

spouts/spout.php
```diff
@@ -91,14 +91,16 @@ public function getSpoutTitle() {
     abstract public function getId();
 
     /**
-     * returns the current title as string
+     * returns the current title as string with html special chars decoded if
+     * applicable.
```
jtojnar: When does `applicable` occur?

spouts/rss/feed.php
niol: Because some people wanted HTML in the title to be shown.

jtojnar: Why is it escaped for GitHub spout then?

niol: Because otherwise, a commit message containing HTML entities would mess up the selfoss stream in the client browser (or be stripped if the tag is not allowed). What we want from commit messages is the raw HTML to be shown in the browser, thus the escaping.

On the contrary, in RSS feeds, we want the HTML to format text in the client browser. Thus, it must be passed as raw HTML the the content loader, filtered by sanitize{Field,Content}() and sent the the client browser.

jtojnar: [RSS specs](https://cyber.harvard.edu/rss/rss.html#hrelementsOfLtitemgt) suggest HTML can only appear within `description` element. The following feed was broken by the PR:

http://www.vzhurudolu.cz/rss

niol: Are you sure? From a quick test, the behavior for this feed is the same as before, which was my intention. A few years ago, some people were expecting HTML in the title. No sure what to do here.

jtojnar: You are right, the behaviour was the same. And apparently, Feedly and tt-rss do it too. For now we can probably leave it as is

PR #907, opened 2017-03-26
make it unlikely that the client browser gets outdated js or css

niol:
This change makes the js and css generated to a file containing the
max mtime of source material in its name. Therefore, each time the
source js or css files are changed, a new file with a new name gets
generated and referenced in the app. From the client browser's view,
this is a new ressource that needs to get fetched regardless of the
cache status of older css or js.

The trade-off is that each request implies stat'ing each source
material file. There is no notiecable time penalty in loading time.

### DIFF
--- a/helpers/View.php
+++ b/helpers/View.php
@@ -13,11 +13,23 @@ class View {
     /** @var string current base url */
     public $base = '';
 
+    /** @internal JS static resource type */
+    const STATIC_RESOURCE_JS = 'js';
+    /** @internal CSS static resource type */
+    const STATIC_RESOURCE_CSS = 'css';
+
+    public static $staticmtime = [
+        self::STATIC_RESOURCE_JS => 0,
+        self::STATIC_RESOURCE_CSS => 0
+    ];
+    public static $staticPrefix = 'all';
+
     /**
      * set global view vars
      */
     public function __construct() {
-        $this->genMinifiedJsAndCss();
+        $this->genMinified(self::STATIC_RESOURCE_JS);
+        $this->genMinified(self::STATIC_RESOURCE_CSS);
         $this->base = $this->getBaseUrl();
     }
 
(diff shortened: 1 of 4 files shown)


helpers/View.php
```diff
@@ -139,33 +159,49 @@ public static function getGlobalJsFileName() {
...
-        // minify js
-        $targetJs = \F3::get('BASEDIR') . '/public/' . self::getGlobalJsFileName();
-        if (!file_exists($targetJs) || \F3::get('DEBUG') != 0) {
-            $js = '';
-            foreach (\F3::get('js') as $file) {
-                $js = $js . "\n" . $this->minifyJs(file_get_contents(\F3::get('BASEDIR') . '/' . $file));
+    private function genMinified($type) {
+        self::$staticmtime[$type] = self::maxmtime(\F3::get($type));
+
+        if ($type == 'js') {
```
jtojnar: Rather than 'js' and 'css', class constants should be created to decrease the chance of accidental mistypes. Since older versions of PHP do no support private constants, `@internal` annotations should be used for now.

helpers/View.php
```diff
@@ -125,12 +129,28 @@ public function jsonSuccess($data) {
     }
 
     /**
+     * returns max mtime for file paths given in array
+     *
+     * @param filePaths array of file paths
+     * @returns max mtime as int (unix timestamp)
+     */
+    public static function maxmtime($filePaths) {
```
jtojnar: `array` type hint should be added since it is supported since PHP 5.1: https://3v4l.org/U4Gq4

helpers/View.php
```diff
@@ -139,33 +159,49 @@ public static function getGlobalJsFileName() {
...
         }
 
-        // minify css
-        $targetCss = \F3::get('BASEDIR') . '/public/' . self::getGlobalCssFileName();
-        if (!file_exists($targetCss) || \F3::get('DEBUG') != 0) {
-            $css = '';
-            foreach (\F3::get('css') as $file) {
-                $css = $css . "\n" . $this->minifyCss(file_get_contents(\F3::get('BASEDIR') . '/' . $file));
+        // cleanup
+        if ($builtSomething) {
```
jtojnar: Why not just use cache busting using query string?

niol: One disadvantage of cache busting using a query string is that it does not return a 404 for outdated content. But is saves the cleanup. Anyway, both may lower user support needs.

jtojnar:
Website should be updated as well.

PR #915, opened 2017-04-13
Fix youtube spout

Muges:
This pull request fixes the youtube spout to make it work with the following urls :
 - `https://www.youtube.com/channel/<channel-id>` where `<channel-id>` contains an hyphen;
 - `https://www.youtube.com/user/<username>`.

### DIFF
--- a/spouts/youtube/youtube.php
+++ b/spouts/youtube/youtube.php
@@ -49,9 +49,12 @@ class youtube extends \spouts\rss\feed {
      */
     public function getXmlUrl($params) {
         $channel = $params['channel'];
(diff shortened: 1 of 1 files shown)


jtojnar:
Thanks for the pull request. It will be included in 2.18.

jtojnar:
Hi @Muges. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions under `GPL-3-or-later`?

PR #916, opened 2017-04-14
[WIP] Refactoring

jtojnar:
I have  been trying to use `null` instead of `false` or `''` whenever possible. Additionally, I also changed majority of comparisons to strict version, filled in missing parameters to phpdoc and also added `array` type hints.

**Tests should be merged first.**

### DIFF
--- a/spouts/youtube/youtube.php
+++ b/spouts/youtube/youtube.php
@@ -91,8 +91,8 @@ class youtube extends \spouts\rss\feed {
                 return @$item->get_enclosure(0)->get_link();
             }
         } else { // no enclosures: search image link in content
-            $image = $this->getImage(@$item->get_content());
-            if ($image !== false) {
+            $image = \helpers\Image::findFirstImageSource(@$item->get_content());
+            if ($image !== null) {
                 return $image;
             }
         }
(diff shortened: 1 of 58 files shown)


spouts/youtube/youtube.php
```diff
@@ -91,11 +91,37 @@ public function getThumbnail() {
...
+     * taken from: http://zytzagoo.net/blog/2008/01/23/extracting-images-from-html-using-regular-expressions/
+     * Searches for the first occurence of an html <img> element in a string
+     * and extracts the src if it finds it. Returns null in case an <img>
+     * element is not found.
+     *
+     * @param string $html An HTML string
+     *
+     * @return ?string content of the src attribute of the first image
+     */
+    private function getImage($html) {
```
jtojnar: factor out to a helpers class

jtojnar:
For some reason https://feeds.feedburner.com/sandraandwoo now uses feedburner favicon instead of the site one.

niol:
Yes this happens since merging #920 .

jtojnar:
You are right, I was probably using old master on my selfoss instance.

PR #917, opened 2017-04-14
fix no favicon for rss feeds

niol:


spouts/rss/feed.php
```diff
@@ -243,7 +243,7 @@ public function getContent() {
      * @return string icon url
      */
     public function getIcon() {
-        if (isset($this->faviconUrl)) {
+        if ($this->faviconUrl !== '') {
```
jtojnar: Would not it be better to set it to null in the initial definition?

niol: I have no argument for or against this. The test must match the initial definition, otherwise it doesn't work, that's all.

jtojnar: Then I would prefer that. I would like to have as much type checking as possible in PHP.

Issue #925, opened 2017-05-09
Feature Request: start with Tag


aslmx:
It is possible to configure selfoss to start with unread, newest or starred.

How about a setting that allows which tag (or source) is used to start with?

It does not necessarily be a new setting, it could be implemented into the existing one. If it is not newest, starred or unread, then assume to show unread for the specified tag or if the tag doesn't exist, show unread for source. If source does not exist then default to whatever the current default is (i think it is newest).

Of course a new setting would make it most flexible and adaptable to users wishes (e.g. show starred articles for feed / $tag)

jtojnar:
If the auth form remembered the path, the problem could be solved by bookmarking `https://yourselfossinstance.com/#unread/tag-xyz` and accessing that. This could also help with #254.

niol:
The server does not know the hash, so the auth form would have to be posted asynchronously.

Anyway, the following and some doc updates should do the trick I think:

```diff
diff --git a/public/js/selfoss-base.js b/public/js/selfoss-base.js
index caf16f1..02afa92 100644
--- a/public/js/selfoss-base.js
+++ b/public/js/selfoss-base.js
@@ -59,9 +59,6 @@ var selfoss = {
             // set items per page
             selfoss.filter.itemsPerPage = $('#config').data('items_perpage');

-            // initialize type by homepage config param
-            selfoss.filter.type = $('#config').data('homepage');
-
             // read the html title configured
             selfoss.htmlTitle = $('#config').data('html_title')

diff --git a/public/js/selfoss-events.js b/public/js/selfoss-events.js
index ab5ad11..e5594ca 100644
--- a/public/js/selfoss-events.js
+++ b/public/js/selfoss-events.js
@@ -35,8 +35,11 @@ selfoss.events = {
         $(window).bind("resize", selfoss.events.resize);
         selfoss.events.resize();

-        if( location.hash == '' )
-            selfoss.events.setHash($('#config').data('homepage'), 'all');
+        if( location.hash == '' ) {
+            var homePagePath = $('#config').data('homepage').split('/');
+            if (!homePagePath[1]) homePagePath.push('all');
+            selfoss.events.setHash(homePagePath[0], homePagePath[1]);
+        }

         // hash change event
         window.onhashchange = selfoss.events.hashChange;
```

jtojnar:
Well, the hash could be stored inside a hidden form field or the auth form could be rendered client-side with AJAX authentication. The latter will be needed for decoupling the JS client from the server anyway.

[...]

PR #927, opened 2017-05-10
'homepage' config: make it possible to start with a tag or a source (#925)

niol:


jtojnar:
Looks good.

One minor thing, if I set `homepage=sources` and then try to switch to a source, it triggers “Invalid section: sources”. Even if it is not described in documentation, some users still might try it. Though it probably is not worth of addressing with the current routing system.

niol:
I think the error message is enough otherwise the validation needs to be duplicated in the conf validation.

Issue #919, opened 2017-04-18
Deauthentication not recognised by the client


jtojnar:
1. Log in to selfoss
2. Delete the `PHPSESSID` cookie
3. Switch the tag or something, the following error will be produced:

> Load list error: parsererror SyntaxError: JSON.parse: unexpected character at line 1 column 1 of the JSON data

Alternately, for 3., let selfoss synchronize the data, then a more precise error will be shown:

> Could not sync last changes from server: error Forbidden

Weishaupt:
I guess I experience a similar problem. On multiple clients (using firefox, chrome mobile) I only log in once. When I reopen the page, no login dialogue is shown and the error reads

> List load error: error Forbidden

If I'm using a private session, the login page is shown, and the client properly loads the feed items.

I'm on master.

jtojnar:
Yes, it was made worse [here](https://github.com/SSilence/selfoss/pull/931#discussion_r120657546). I should have requested for it to be fixed, did not realize it would be so annoying.

virusmoere:
Still not working for me on the current dev version.

niol:
Works for me. Can you detail how you make it fail and the corresponding error message?

virusmoere:
It is enough to delete the PHPSESSID cookie. After that I get the "No items loaded, reload?" message on the site and in the dev tools I can see calls on "https://url/?offset=0&itemsPerPage=50&search=&type=unread&tag=&source=&sourcesNav=false&ajax=true" that are answered with 403 unautorized - "an error occured".

virusmoere:
Not for me:
![selfoss](https://user-images.githubusercontent.com/6796393/41842442-10c4d0a0-786b-11e8-89e5-cd06770a07f5.PNG)

My config.ini for reference:
`; see https://selfoss.aditu.de for more information about
; the configuration parameters
[globals]
db_type=sqlite
db_file=data/sqlite/selfoss.db
db_host=localhost
db_database=selfoss
db_username=root
db_password=
db_port=
db_prefix=
logger_destination=file:data/logs/default.log
logger_level=ERROR
items_perpage=50
items_lifetime=30
base_url=https://***
username=***
password=***
salt=***
public=
html_title=selfoss
rss_title=selfoss feed
rss_max_items=300
rss_mark_as_read=0
homepage=unread
language=0
auto_mark_as_read=1
auto_collapse=0
auto_stream_more=1
anonymizer=
use_system_font=
share=gtfpde
wallabag=
[...]

PR #931, opened 2017-05-12
Improve js client

niol:
Some minor enhancements and refactoring, use API for login and update item datetimes client-side.

### DIFF
--- a/controllers/Index.php
+++ b/controllers/Index.php
@@ -229,8 +229,8 @@ class Index extends BaseController {
         $itemsHtml = '';
 
         $firstPage = $options['offset'] == 0
-            && $options['offset_from_id'] == ''
-            && $options['offset_from_datetime'] == '';
+            && $options['fromId'] == ''
+            && $options['fromDatetime'] == '';
         if ($options['source'] && $this->allowedToUpdate() && $firstPage) {
             $itemsHtml = '<button type="button" id="refresh-source" class="refresh-source">' . \F3::get('lang_source_refresh') . '</button>';
         }
--- a/controllers/Index.php
+++ b/controllers/Index.php
@@ -102,7 +102,6 @@ class Index extends BaseController {
      * @return void
      */
     public function password() {
-        $this->view = new \helpers\View();
         $this->view->password = true;
         if (isset($_POST['password'])) {
             $this->view->hash = hash('sha512', \F3::get('salt') . $_POST['password']);
@@ -117,8 +116,6 @@ class Index extends BaseController {
      * @return void
      */
     public function login() {
-        $view = new \helpers\View();
-
         $error = null;
 
         if (isset($_REQUEST['username'])) {
@@ -136,19 +133,19 @@ class Index extends BaseController {
         }
 
         if ($error !== null) {
-            $view->jsonError([
+            $this->view->jsonError([
                 'success' => false,
                 'error' => $error
             ]);
         }
 
         if (\F3::get('auth')->login($username, $password)) {
-            $view->jsonSuccess([
+            $this->view->jsonSuccess([
                 'success' => true
             ]);
         }
 
-        $view->jsonSuccess([
+        $this->view->jsonSuccess([
             'success' => false,
             'error' => 'invalid username/password'
         ]);
@@ -161,9 +158,8 @@ class Index extends BaseController {
      * @return void
      */
     public function logout() {
-        $view = new \helpers\View();
         \F3::get('auth')->logout();
-        $view->jsonSuccess([
+        $this->view->jsonSuccess([
             'success' => true
         ]);
     }
--- a/controllers/Index.php
+++ b/controllers/Index.php
@@ -224,7 +224,6 @@ class Index extends BaseController {
      * @return string html with items
      */
     private function loadItems($options, $tags) {
-        $tagColors = $this->convertTagsToAssocArray($tags);
         $itemDao = new \daos\Items();
         $itemsHtml = '';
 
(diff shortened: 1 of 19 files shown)


public/js/selfoss-db.js
```diff
@@ -1,6 +1,154 @@
 /**
  * db functions: client data repository
  */
+
+
+selfoss.dbOnline = {
```
jtojnar: I do not understand the name. Under database, I usually imagine something that stores data.

controllers/Index.php
```diff
@@ -112,52 +111,36 @@ public function password() {
...
+
+        if (isset($_REQUEST['password'])) {
+            $password = $_REQUEST['password'];
+        } else {
+            $password = '';
+            $error = 'no password given';
+        }
+
+        if ($error !== null) {
+            $view->jsonSuccess([
```
jtojnar: Why do we use `jsonSuccess` instead of `jsonError`?

public/js/selfoss-ui.js
```diff
@@ -4,6 +4,60 @@
 selfoss.ui = {
 
 
+    showLogin: function(error) {
+        var error = (typeof error !== 'undefined') ? error : '';
```
jtojnar: Maybe taking care of this in css would be simpler:

```css
#nav-refresh, … {
    display: none;
}
body.publicmode #nav-refresh, body.loggedin #nav-refresh, … {
    display: block;
}
```

public/js/selfoss-base.js
```diff
@@ -75,7 +90,9 @@ var selfoss = {
             window.setInterval(selfoss.dbOnline.sync, 60*1000);
 
             window.setInterval(selfoss.ui.refreshEntryDatetimes, 60*1000);
-        });
+
+            selfoss.ui.showMainUi();
```
jtojnar: The UI should be shown before the events are initialised otherwise events.resize will be fired prematurely and the custom scroll bar will be calculated incorrectly:

public/js/selfoss-db.js
niol: Here it fetches data. I'm open to other propositions.

controllers/Index.php
niol: It made not much sense but the login API was like that before so I did ont change et.

jtojnar: Okay, I will take care of this in the refactoring pull request.

public/js/selfoss-db.js
jtojnar: Was there any particular motivation for moving it into a separate file?

The main issue I have with this is probably the entangledness of the architecture, though refactoring it is a long term goal.

niol: It makes the base file a little less long and makes sense regarding my work in https://github.com/niol/selfoss/blob/ppr/offline/public/js/selfoss-db.js

controllers/Index.php
niol: jsonError adds a ```400 Bad Request``` error. If password or username are missing, it is a bad request, so you are right, I'll fix this. This is not the case for the auth error.

public/js/selfoss-base.js
jtojnar: I can still reproduce this.

public/js/selfoss-db.js
```diff
@@ -1,6 +1,154 @@
...
+            selfoss.lastUpdate = new Date(0);
+            getStatuses = undefined;
+        }
+
+        return $.ajax({
+            url: 'items/sync',
+            type: 'GET',
+            dataType: 'json',
+            data: {
+                since:         selfoss.db.lastUpdate.toISOString(),
```
jtojnar: I am getting `TypeError: selfoss.db.lastUpdate is undefined` here.

templates/home.phtml
```diff
@@ -52,7 +52,7 @@
 </head>
 <body class="
     <?= $this->publicMode === true ? 'publicmode' : ''; ?>
-    <?= $this->loggedin === true ? 'loggedin' : 'notloggedin'; ?>
+    <?= $this->authEnabled === true ? 'auth' : ''; ?>
```
jtojnar: I am still getting the auth form even when the authentication is disabled.

public/js/selfoss-base.js
niol: I cannot manage to reproduce. Can you please describe the steps?

jtojnar: After you log in it seems to work fine but if you refresh the page, it overflows. You need to have enough tags or items to fill the whole area, or resize the window before reloading the page. I can reproduce it in both Firefox and Chomium.

public/js/selfoss-db.js
jtojnar: As I see it, these three objects form the model layer. Optimally, the `dbOnline` object should only take care of fetching/uploading data from the server. It could then be called `synchronizer`. The `dbOffline` should only be a nice wrapper above `dexie`, I would call it `storage`. Last, the `db` object could be an envelope above `synchronizer` and `storage`, providing an opaque model for the rest of the app. What I find the most important, none of the object should have any access to DOM.

templates/login.phtml
```diff
@@ -37,20 +37,6 @@
 </head>
```
jtojnar: This file can be renamed.

public/js/selfoss-db.js
niol: Okay then you can rebase this patch out later.

public/js/selfoss-base.js
```diff
@@ -97,6 +118,75 @@ var selfoss = {
...
+
+
+    clearSession: function() {
+        Cookies.remove('onlineSession', {path: window.location.pathname});
+        selfoss.loggedin = false;
+    },
+
+
+    hasSession: function() {
+        selfoss.loggedin = Cookies.get('onlineSession') == 'true';
```
niol: This cookie is a hint that there was an online session when last loaded, and will be removed at startup when PHPSESSID is cleared. When that PHPSESSID is cleared, you get like before the 403 error after each ajax request. The next step is to fix #919 .

public/js/selfoss-base.js
```diff
@@ -97,6 +118,75 @@ var selfoss = {
...
+            complete: function(jqXHR, textStatus) {
+                $('#loginform').removeClass('loading');
+            }
+        });
+        e.preventDefault();
+    },
+
+
+    logout: function() {
+        selfoss.clearSession();
```
niol: Again, this is all done with offline in mind and is only a hint that there was an online session last time.

PR #934, opened 2017-06-01
Improvements of Twitter spout

jtojnar:


jtojnar:
Fancybox does not work for some reason when the image URL does not end with image extension 🤷‍

jtojnar:
Fancybox is now fixed, tests will be added later when testing framework is chosen.

PR #943, opened 2017-06-09
Make source code checkers accessible from Grunt

jtojnar:


jtojnar:
I do not really like this. I [asked](https://www.reddit.com/r/PHP/comments/6hg7dx/how_do_you_handle_incompatible_dev_dependencies/) on reddit and so far we have two alternatives:

* [composer-tool-installer-plugin](https://github.com/tommy-muehle/composer-tool-installer-plugin)
* [composer-bin-plugin](https://github.com/bamarni/composer-bin-plugin)

jtojnar:
In the end, I have decided to make users install the linters manually, either using a package manager or by calling `utils/install-phars.sh`.

PR #944, opened 2017-06-14
Log database driver name when estabilishing a connection

jtojnar:


jtojnar:
I do not really like this, perhaps we should leave it to some DB wrapper.

PR #949, opened 2017-06-16
Fix fancybox being considered extraneous package

jtojnar:
Since Debian-provided version of NPM does not support scoped packages, we have to install FancyBox from Git. I also changed the dependency name in the `package.json` to just `fancybox` in order
not to trigger the outdated NPM. Apparently, this will make NPM think the package is extraneous due to a mismatch between the package name and the name in the `package.json`.

### DIFF
--- a/public/package.json
+++ b/public/package.json
@@ -1,6 +1,6 @@
 {
   "dependencies": {
-    "fancybox": "github:fancyapps/fancybox#v3.1.20",
+    "@fancyapps/fancybox": "^3.1.20",
     "jquery": "^2.2.4",
     "jquery-hotkeys": "^0.2.2",
     "jquery-mousewheel": "^3.1.13",


kolydart:
I get the following error on `cd public; npm install;`

> npm http 304 https://registry.npmjs.org/fancybox
> npm ERR! Error: No compatible version found: fancybox@'github:fancyapps/fancybox#v3.1.20'
> npm ERR! Valid install targets:
> npm ERR! ["2.1.5","2.1.6","2.1.7","2.1.8","3.0.0"]
> npm ERR!     at installTargetsError (/usr/share/npm/lib/cache.js:719:10)

ubuntu server 14.04 lts 
default LAMP install

kolydart:
Actually, the error happens on current master commit (08d5f15)

jtojnar:
I tried it on 14.10 VM (npm 1.4.21) and it indeed fails with the error. I tried various permutations including the `fancyapps/fancybox`, which should work [since 1.1.65](https://stackoverflow.com/questions/17509669/how-to-install-an-npm-package-from-github-directly#comment40939211_21918559) but everything fails.

Since @olsonpm remains adamant about [transferring the `fancybox` package](https://github.com/olsonpm/fancybox/issues/3), we will have to convince the NPM folks. Hopefully, they will recognize it does not make sense to treat non-semver package like semver one.

jtojnar:
I am unable to make it work so we have just decided to ignore the old NPM for now.

PR #952, opened 2017-06-22
proposal: consistent styling for message action and stream buttons

niol:
This restyling of the stream buttons and message action button has been working quite well for me.
screenshot

### DIFF
--- a/public/css/style.css
+++ b/public/css/style.css
@@ -49,8 +49,12 @@ body * {
     font-size:0.9em;
     text-align:center;
     border-radius:2px;
+    padding:5px;
+    padding-left: 2em;
+    padding-right: 2em;
     cursor:pointer;
-    background:#cccccc;
+    color:#d7d7d7;
+    background:#3d6d69;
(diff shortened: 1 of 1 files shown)


jtojnar:
I actually liked the message action button having the same colour as the message bar, even GTK [does it](https://developer.gnome.org/hig/stable/info-bars.html.en).

Either way,

1. The dark turquoise colour is special and should not be overused.
2. The more button is not really important – most of the time scrolling will suffice. The colour should be reserved for important buttons like “log in”.
3. It does not really look well on light backgrounds.

niol:
Okay for the message action button.
Maybe it makes more sense for the stream buttons with ```auto-stream_more=0``` in mobile view.

screenshot

jtojnar:
It looks slightly better than the current one:



I still do not like the colour on a light background, maybe something lighter:



Or even more de-saturated:



Either way, localizations should be considered, 150px might be to narrow for many languages.

PR #974, opened 2017-08-08
Implement a spout tracking GitHub tags.

bertptrs:
Add a new spout that tracks new tags on GitHub projects. It has support for consistent IDs and dates.

Unfortunately, the GitHub API does not include dates in the tags request, so an additional request is needed to get the date. To work around this, I included a tweakable setting that limits the number of tags retrieved at once, so that the impact of this limitation is reduced.

I formatted the sources according to your `.php_cs` file, let me know if I missed anything. I did not use the `WebClient::getHttpClient` method since it is not available in the 2.17 release and I really wanted to use this myself.

### DIFF
--- a/spouts/github/tags.php
+++ b/spouts/github/tags.php
@@ -1,0 +1,174 @@
+<?php
+
+namespace spouts\github;
+
+use helpers\WebClient;
+
+/**
+ * Spout for tracking tags on GitHub.
+ *
+ * This spout is loosely based on the "commits" spout.
+ *
+ * @copyright Copyright Bert Peters (http://bertptrs.nl)
+ * @license GPLv3 (https://www.gnu.org/licenses/gpl-3.0.html)
+ * @author Bert Peters <bert@bertptrs.nl>
+ */
+class tags extends commits {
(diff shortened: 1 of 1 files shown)


jtojnar:
Is there any difference from [Tags RSS feed](https://github.com/SSilence/selfoss/tags.atom) other than the links to the tarball? You can get the links + a changelog when using the feed with FullTextRSS spout.

In the future I want to add better feed auto-detection – e.g. by entering the https://github.com/SSilence/selfoss URL, user would be offered the following options:

* Commits: https://github.com/SSilence/selfoss/commits.atom
* Tags: https://github.com/SSilence/selfoss/tags.atom
* Releases: https://github.com/SSilence/selfoss/releases.atom

but for now, adding the URL manually be enough.

bertptrs:
I don't think there is much of a difference, I added it mostly because I already saw the "Github" spout and did not think to look any further.

jtojnar:
We will probably migrate away from the existing GitHub spout in the future. There is no point in maintaining a custom spout when RSS exists.

PR #977, opened 2017-08-17
Improve error messages

jtojnar:
Currently when an exception is raised, we will only see “an error occurred” message. This is not very useful since it provides no instructions on what to do.

Instead we should point user to check the logs or even point out that the logs are read-only when that is the case.


jtojnar:
I think we should extend this to even more common errors like incorrect database and config permissions. Also it would be nice to have 500 show an error message to client so that they do not need to check the console/logs. Though we need to make sure that the backtrace is only dumped for authenticated users for security reasons.

PR #981, opened 2017-09-05
Update of the FRENCH translation for selfoss

Galixte:
Hi,

could you merge this, please?

It could be interresting to can translate the error messages, like: 
- “invalid username/password”,
- “2017-09-05 20:00:48 error loading feed content: http://www.domain.tld/feed.php is invalid XML, likely due to invalid characters. XML error: Mismatched tag at line 51, column 8”,
- “2017-09-05 20:01:03 error loading feed content: cURL error 22: The requested URL returned error: 404”.

Moreover some field appear but these aren’t translatable in language file :
- for ```Deviant art - favs of a user```, ```Deviant art - user``` & ```tumblr User``` are selected, the field “Username”,
- for ```Facebook page feed``` is selected, the field: “Page name”,
- for ```GitHub``` is selected, the field: 
  - “Owner”,
  - “Repository”,
  - “Branch”,
- for ```News: Golem```, ```News: Heise``` or ```News: Pro-Linux``` are selected, the field: “Section”,
- for ``Reddit``` is selected, the field: 
  - “Subreddit or multireddit url”,
  - “Username”,
  - “Password”,
- for ```RSS Feed …``` is selected, the field: “URL”,
- for ```RSS Feed (with readability)``` is selected, the field: “Readability API Key”,
- for ```Twitter …``` is selected, the field: 
  - “Consumer Key”,
  - “Consumer Secret”,
  - “Access Token (optional)”,
  - “Access Token Secret (optional)”,
  - “List Slug”,
  - “Username”,
  - “Access Key”,
  - “Access Secret”,
- for ```YouTube Channel``` is selected, the field: “Channel.


fmunch:
I agree with most of your changes, but I have a few comments:

- `lang_newest`: imho "Récents" better describes the fact that not every entry is kept, only the most recent ones.
- `lang_mark` and `lang_unmark`: "Marquer comme lu" et "Marquer comme non lu" will be too wide for the mobile UI. The first one pushes "Charger les images" under the close button, and the second one makes the toolbar appear on two lines.
- `lang_markread`: the better choice would probably be "Marquer comme lu" (or "Marquer tout comme lu"), but it would be too wide. "Noter" does not really have the same meaning, more like "making the entry stand out"? Not sure it would be better than "Marquer lu", even if it allows to add "comme".

Galixte:
Hi @fmunch 

> lang_newest: imho "Récents" better describes the fact that not every entry is kept, only the most recent ones.

I changed « Récents » with « Tous » because solfoss propose me all feeds stored in database when i click on this button. I set ```items_lifetime=360```, so i have access to 1 year of feeds, it’s not recent it corresponds to 360 days, so it’s a non-sense to name the button « Recents ».

> lang_mark and lang_unmark: "Marquer comme lu" et "Marquer comme non lu" will be too wide for the mobile UI. The first one pushes "Charger les images" under the close button, and the second one makes the toolbar appear on two lines.

Ok, but it’s better to invert the text, because there isn’t the ```title="Marquer comme lu"``` to explain : « This is the button to mark as read », IMHO it should be.

> lang_markread: the better choice would probably be "Marquer comme lu" (or "Marquer tout comme lu"), but it would be too wide. "Noter" does not really have the same meaning, more like "making the entry stand out"? Not sure it would be better than "Marquer lu", even if it allows to add "comme".

« Marquer comme lu » appears on two lines, that’s why i proposed « Noter » it’t the shortest synonym word i found. « Marquer lu » it’s a joke. :) 
« Marquer tout comme lu » takes place in two line but at least the two lines are fulfilled.

jtojnar:
Thanks for the pull request.

I have opened #985 which added `lang_login_invalid_credentials` for the first error. Unfortunately, the two other errors are not so easy to translate as they are basically raw text message returned by some underlying library. I filed an issue for that #986.

Regarding spout form fields, they could be also translated but I need to decide how to best do that, especially with regards to third-party spouts. I started #987.

As for the translation itself:

* I agree that the filter name “Recent” filter is quite unfortunate. All would be better and I intend to replace it in English too.
* This will be definitely confusing either way. If there is no shorter variant, the longer should be used and the user interface edited so the text would fit.
* Two line text is fine here.

Galixte:
Hi @jtojnar i updated the translation, could you merge this, please?

jtojnar:
Merged as b8f51c11a7ed92eb4b88cbc983bbbc1a47103d23, thanks.

jtojnar:
Hi @Galixte. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contribution under `GPL-3-or-later`?

Galixte:
> Hi @Galixte. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contribution under `GPL-3-or-later`?
> 

Sorry for the late, yes I’m agree with that.

Issue #967, opened 2017-07-21
cliupdate sempahore locker


alexalouit:
Hi,

I found a issue (in my opinion).
Yesterday, I found my database in trouble with many connections and simultaneous writings.
I noticed that there were ~30 cron processes (cliupdate) running.

I think a semaphore would have to be set up to prevent this.

niol:
Yes I had done that in #597 but this was not merged.

jtojnar:
What are the merits of locking, compared to `UNIQUE` constraint (#921)? At a glance, the constraint [seems cleaner](https://github.com/SSilence/selfoss/issues/208#issuecomment-296114446) but I did not give it a proper thought yet.

alexalouit:
I think file locker will not do useless query, is upper in application level.
Nevertheless, the two systems remain complementary.

This can prevent process &nd memory bombing from cron zombie/defunct processes.

jtojnar:
I have originally intended to update the items in the races (hence the `UPSERT` mention) but since the probability of items changing between the races is low, the queries would indeed be useless. With that in mind, I can see the complementarity; I am not convinced that this is not a premature optimization not worth the added complexity, though.

alexalouit:
```ALTER IGNORE TABLE `items` ADD UNIQUE INDEX (`uid` (150), `source`);```

from  ~111 619 to ~67 562 entries.

alexalouit:
Same error this morning, multiple process cliupdate performs hugely of sql query (which are denied from the unique key).

niol:
I think that any running update should make another one exit immediately to prevent both sources hammering and rogue processes. There are multiple ways of implementing this, I have proposed using file locks, another way would be to use unix sockets (not portable) or the database. I'll simplify my patch and submit a pull request if it has some chance of getting accepted. Otherwise, this issue should be closed.

jtojnar:
One might also argue that locking should not be a concern of the application and be done on a higher level:

* If I understand correctly, systemd timers only allow one instance of the service running, at least [`Persistent` option](http://man7.org/linux/man-pages/man5/systemd.timer.5.html#OPTIONS) hints so.
* For traditional UNIX, `flock` or similar command can be [used with cron](https://serverfault.com/questions/82857/prevent-duplicate-cron-jobs-running).
* For shared web hosts that do not allow running UNIX commands, creating an optional PHP wrapper simulating `flock` could be used.

[...]

PR #991, opened 2017-11-02
bugfix & improvements

alexalouit:
add semaphore locker
add cli timeout
add db unique key

### DIFF
--- a/cliupdate.php
+++ b/cliupdate.php
@@ -1,4 +1,5 @@
 <?php
+set_time_limit(300);
 
 chdir(__DIR__);
 require __DIR__ . '/common.php';
--- a/cliupdate.php
+++ b/cliupdate.php
@@ -1,6 +1,4 @@
 <?php
-set_time_limit(300);
-
 chdir(__DIR__);
 require __DIR__ . '/common.php';
 
--- a/daos/sqlite/Database.php
+++ b/daos/sqlite/Database.php
@@ -234,6 +234,47 @@ class Database {
                         'INSERT INTO version (version) VALUES (11)'
                     ]);
                 }
+                if (strnatcmp($version, '12') < 0) {
+                    \F3::get('db')->exec([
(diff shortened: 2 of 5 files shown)


cliupdate.php
```diff
@@ -1,4 +1,5 @@
 <?php
+set_time_limit(300);
```
jtojnar: Why this time limit? We are already using a greater one in the `ContentLoader`.

alexalouit:
```SOURCE_UPDATE_MAX_INTERVAL``` fix in 4726c4c29cbc8bd4393f2d58f02c9c60734f4543

File: daos/sqlite/Database.php [hunk 2]
```diff
@@ -234,6 +234,47 @@ public function __construct() {
                         'INSERT INTO version (version) VALUES (11)'
                     ]);
                 }
+                if (strnatcmp($version, '12') < 0) {
+                    \F3::get('db')->exec([
+                        // Table needs to be re-created because ALTER TABLE is rather limited.
```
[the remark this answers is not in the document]
alexalouit: It is better? 2cf87cad222a46cbfe9756869f5ef2f5bee9b2f8

cliupdate.php
alexalouit: fix in 54235f84fd374354beb3e80e68d3ecca0590206d

alexalouit:
query fixed in 54235f84fd374354beb3e80e68d3ecca0590206d

jtojnar:
Regarding the locking, in https://github.com/SSilence/selfoss/issues/967#issuecomment-319159139, I actually meant the `flock` command as demonstrated in https://serverfault.com/questions/82857/prevent-duplicate-cron-jobs-running. On selfoss site it would mean creating `cliupdate_flock.php` that would allow only one instance of itself at a time.

jtojnar:
Also it would be nice if you could split the unique constraint to a separate commit from the locking part.

alexalouit:
see #994 and #995

PR #994, opened 2017-11-03
add locker

alexalouit:
per source file locker

### DIFF
--- a/helpers/ContentLoader.php
+++ b/helpers/ContentLoader.php
@@ -2,6 +2,8 @@
 
 namespace helpers;
 
+define('SOURCE_UPDATE_MAX_INTERVAL', 300);
+
 /**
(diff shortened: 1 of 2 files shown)


jtojnar:
This is basically #597, which was previously rejected by Tobias. In https://github.com/SSilence/selfoss/issues/967#issuecomment-319159139, I actually meant the `flock` command as demonstrated in https://serverfault.com/questions/82857/prevent-duplicate-cron-jobs-running. Emulating it on selfoss side would mean creating `cliupdate_flock.php` that would allow only one instance of itself at a time.

alexalouit:
That's makes sense.

There remain two solutions, the basic use of ascii file without locking, or a field (timestamp or boolean) in database.
I think second is much better.

PR #995, opened 2017-11-03
Add unique key

alexalouit:
unique key for mysql
unique key for sqlite

### DIFF
--- a/src/daos/mysql/Database.php
+++ b/src/daos/mysql/Database.php
@@ -231,6 +231,12 @@ class Database implements \daos\DatabaseInterface {
                     'INSERT INTO ' . \F3::get('db_prefix') . 'version (version) VALUES (12)'
                 ]);
             }
+            if (strnatcmp($version, '13') < 0) {
+                $this->exec([
+                    'ALTER TABLE ' . \F3::get('db_prefix') . 'items ADD UNIQUE INDEX (uid (191), source);',
+                    'INSERT INTO ' . \F3::get('db_prefix') . 'version (version) VALUES (13)'
+                ]);
+            }
         }
     }
 
--- a/src/daos/pgsql/Database.php
+++ b/src/daos/pgsql/Database.php
@@ -219,6 +219,12 @@ class Database implements \daos\DatabaseInterface {
                     'INSERT INTO version (version) VALUES (12)'
                 ]);
             }
+            if (strnatcmp($version, '13') < 0) {
+                $this->exec([
+                    'ALTER TABLE items ADD UNIQUE(uid, source);',
+                    'INSERT INTO version (version) VALUES (13)'
+                ]);
+            }
         }
     }
 
--- a/src/daos/sqlite/Database.php
+++ b/src/daos/sqlite/Database.php
@@ -231,6 +231,12 @@ class Database implements \daos\DatabaseInterface {
                     'INSERT INTO version (version) VALUES (11)'
                 ]);
             }
+            if (strnatcmp($version, '13') < 0) {
+                $this->exec([
+                    'CREATE UNIQUE INDEX uid ON items(uid, source);',
+                    'INSERT INTO version (version) VALUES (13)'
+                ]);
+            }
         }
     }
 


alexalouit:
we need to add unique key for pgsql

daos/sqlite/Database.php
```diff
@@ -235,9 +235,10 @@ public function __construct() {
                     ]);
                 }
                 if (strnatcmp($version, '12') < 0) {
-                    \F3::get('db')->exec('
-                        CREATE UNIQUE INDEX uid ON new_items(uid, source);
-                    ');
+                    \F3::get('db')->exec([
+                        'CREATE UNIQUE INDEX uid ON new_items(uid, source);',
+                        'INSERT INTO version (version) VALUES (12)'
```
alexalouit: I'm not sure for the length of index, pgsql support it?

daos/sqlite/Database.php
```diff
@@ -234,6 +234,12 @@ public function __construct() {
                         'INSERT INTO version (version) VALUES (11)'
                     ]);
                 }
+                if (strnatcmp($version, '12') < 0) {
+                    \F3::get('db')->exec([
+                        'CREATE UNIQUE INDEX uid ON new_items(uid, source);',
```
jtojnar: The index should be created on `items`, there should be no table `new_items` at this point.

daos/sqlite/Database.php
```diff
@@ -234,6 +234,12 @@ public function __construct() {
...
+                        'INSERT INTO version (version) VALUES (12)'
```
jtojnar: Could you use version 13 so that it is more clear that the database is current.

daos/pgsql/Database.php
```diff
@@ -227,6 +227,12 @@ public function __construct() {
                         'INSERT INTO version (version) VALUES (12)'
                     ]);
                 }
+                if (strnatcmp($version, '13') < 0) {
+                    \F3::get('db')->exec([
+                        'ALTER TABLE ' . \F3::get('db_prefix') . 'items UNIQUE (uid (150), source);',
+                        'INSERT INTO ' . \F3::get('db_prefix') . 'version (version) VALUES (13)'
```
jtojnar: There is no need to use `db_prefix` in Postgres. It is a measure for cheap shared hostings that only provide a single MySQL database.

daos/pgsql/Database.php
```diff
@@ -227,6 +227,12 @@ public function __construct() {
                         'INSERT INTO version (version) VALUES (12)'
                     ]);
                 }
+                if (strnatcmp($version, '13') < 0) {
+                    \F3::get('db')->exec([
+                        'ALTER TABLE ' . \F3::get('db_prefix') . 'items UNIQUE (uid (150), source);',
```
jtojnar: PostgreSQL is smart enough that no length should be necessary. Additionally, I think the syntax is `ALTER TABLE items ADD UNIQUE(uid, source)`, see the definitions for [`ALTER TABLE`](https://www.postgresql.org/docs/current/static/sql-altertable.html) and [`table_constraint`](https://www.postgresql.org/docs/current/static/sql-createtable.html).

daos/mysql/Database.php
```diff
@@ -243,6 +243,12 @@ public function __construct() {
                         'INSERT INTO ' . \F3::get('db_prefix') . 'version (version) VALUES (12)'
                     ]);
                 }
+                if (strnatcmp($version, '13') < 0) {
+                    \F3::get('db')->exec([
+                        'ALTER IGNORE TABLE ' . \F3::get('db_prefix') . 'items ADD UNIQUE INDEX (uid (150), source);',
```
jtojnar: It looks like `ALTER IGNORE TABLE` syntax is [not supported](https://dev.mysql.com/doc/refman/5.6/en/alter-table.html#idm139885224435136) any more in MySQL 5.7. Manual deduplication will be needed.

jtojnar: Also I am not sure about the 150 byte length. Would not 255 work as well? That should fit 1024 bytes even with four bytes per character.

daos/sqlite/Database.php
alexalouit: 7040883bfbd248dae6c5d266912379c31cfcacd5

daos/sqlite/Database.php
alexalouit: fix in 7040883bfbd248dae6c5d266912379c31cfcacd5

daos/pgsql/Database.php
alexalouit: fix in 7040883bfbd248dae6c5d266912379c31cfcacd5

daos/pgsql/Database.php
alexalouit: fix in 7040883bfbd248dae6c5d266912379c31cfcacd5

daos/mysql/Database.php
alexalouit: fix in 7040883bfbd248dae6c5d266912379c31cfcacd5

alexalouit: On MariaDB 5.5, I got: ```Specified key was too long; max key length is 767 bytes```
It only works with less than ~190.

see https://stackoverflow.com/questions/1814532/1071-specified-key-was-too-long-max-key-length-is-767-bytes/1814594#1814594

jtojnar: Oh, InnoDB. 191 should be used then.

alexalouit: fix in 2cc1cb8

jtojnar:
Rebased this onto master and resolved conflicts.

But yeah, this still needs to handle past duplicates otherwise it will fail with `PDOStatement: UNIQUE constraint failed: items.uid, items.source`

PR #1004, opened 2017-12-06
logging: Allow logging to web server error_log

jtojnar:
This PR adds a new configuration option `logger_destination` allowing to choose between Monolog’s [`ErrorLogHandler`](https://github.com/Seldaek/monolog/blob/master/src/Monolog/Handler/ErrorLogHandler.php) and [`StreamHandler`](https://github.com/Seldaek/monolog/blob/master/src/Monolog/Handler/StreamHandler.php).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,6 +6,7 @@
 - It is possible to set tag or source to be opened after user logs in ([#927](https://github.com/SSilence/selfoss/pull/927))
 - Displaying multiple images from tweets with galleries is supported ([#934](https://github.com/SSilence/selfoss/pull/934))
 - Quoted tweets are supported ([#934](https://github.com/SSilence/selfoss/pull/934))
+- Logging destination can be changed ([#1004](https://github.com/SSilence/selfoss/pull/1004))
 
 ### Bug fixes
 - Fixed Full-text RSS spout ([#897](https://github.com/SSilence/selfoss/pull/897))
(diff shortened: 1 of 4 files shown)


jtojnar:
I am actually not sure if there is any benefit of using `error_log` compared to `file:php://stderr`.

jtojnar:
It looks like the behaviour is slightly different on different servers, however, both go through stderr to server’s error log ([SAPI error log](https://secure.php.net/manual/en/errorfunc.configuration.php#ini.error-log)).

## Apache httpd
Both are written to httpd error log:

* `error_log` function: `[Wed Dec 06 02:12:51.537817 2017] [php7:notice] [pid 20094] [client ::1:49470] [2017-12-06 02:12:51] selfoss.DEBUG: Establish database connection  \n`
* `stderr`: `[2017-12-06 02:25:33] selfoss.DEBUG: Establish database connection`

## nginx + php-fpm
Both are written to php-fpm pool error log (system journal in my case)

* `error_log` function: `Dec 06 03:18:45 srv php-fpm[328]: [WARNING] [pool reader] child 26568 said into stderr: "NOTICE: PHP message: [2017-12-06 03:18:45] selfoss.DEBUG: Establish database connection  "`
* `stderr`: `Dec 06 03:20:43 srv php-fpm[328]: [WARNING] [pool reader] child 26568 said into stderr: "[2017-12-06 03:20:43] selfoss.DEBUG: Establish database connection  "`

It looks like using `stderr` leads to more concise logs.

jantman:
I haven't tested this yet but it looks absolutely wonderful for people like me who are trying to run selfoss in Docker and have existing log management being done at the docker level. This makes it much easier to run selfoss like a proper [12 factor app](https://12factor.net/logs).

Thanks so much!

PR #1010, opened 2018-01-03
fix fr lang_markread

niol:
'Marquer tout comme lu' means 'Mark all read', which is wrong: the button
only marks visible read.


jtojnar:
Original PR for reference https://github.com/SSilence/selfoss/pull/981

PR #1013, opened 2018-01-09
Change icons to vector

jtojnar:
This will make selfoss look crisper on hi-res displays and also allow better `user.css` theming.


jtojnar:
Aargh! With the icons, the generated bundle is huge (2 MB).

I tried subsetting the icons like

```javascript
import { fas } from '@fortawesome/free-solid-svg-icons';
library.add(fas.faCog);
```

and then

```javascript
import { faCog } from '@fortawesome/free-solid-svg-icons';
library.add(faCog);
```

but neither of these really made the bundle smaller.

Finally I switched to [deep imports](https://fontawesome.com/how-to-use/with-the-api/other/tree-shaking#deep-imports)

```javascript
import { faCog } from '@fortawesome/free-solid-svg-icons/faCog';
library.add(faCog);
```

getting the size back down to 0.4 MB.

It turns out Parcel still does not use tree shaking by default, you have to enable it with [`--experimental-scope-hoisting` flag](https://github.com/parcel-bundler/parcel/issues/392#issuecomment-397388573). And enabling it breaks jQuery mousewheel since it is a bowl of spaghetti like all jQuery-related things.

Well never mind, I have been meaning to get rid of the custom scrollbar for a long time. But, apparently, it also [does not support source maps](https://github.com/parcel-bundler/parcel/issues/1564). Welp.

Issue #993, opened 2017-11-02
Create API spec


jtojnar:
We need machine readable API description. Among other things, it will allow us to check the implementation compliance automatically.

I tried to describe the API using OpenAPI 2 but it proved to be too limited. OpenAPI 3 looks more promising but the automated checking tooling does not seem to support it yet.

Depends on: https://github.com/apiaryio/dredd/issues/894

jtojnar:
Bootprint might be more straightforward to set-up than Swagger UI but it also [does not support OpenAPI 3.0](https://github.com/bootprint/bootprint-openapi/issues/91).

```diff
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+/_docs/website/api
 /data/favicons/*.png
 /data/favicons/*.jpg
 /data/thumbnails/*.png
--- a/gruntfile.js
+++ b/gruntfile.js
@@ -163,11 +163,20 @@
         }
     });
 
+    grunt.registerTask('docs:api', 'API documentation', function() {
+        var bootprint = require('bootprint')
+           .load(require('bootprint-openapi'))
+           .build('docs/api-description.json','api')
+           .generate()
+           .done(console.log);
+    });
+
     grunt.registerTask('client:install', 'Install client-side dependencies.', ['auto_install']);
     grunt.registerTask('server:install', 'Install server-side dependencies.', ['composer:install:no-dev:optimize-autoloader:prefer-dist']);
     grunt.registerTask('install', 'Install both client-side and server-side dependencies.', ['client:install', 'server:install']);
     grunt.registerTask('default', ['install', 'versionupdater', 'compress']);
     grunt.registerTask('version', ['versionupdater']);
+    grunt.registerTask('docs', ['docs:api']);
     grunt.registerTask('zip', ['compress']);
     grunt.registerTask('lint:client', 'Check JS syntax', ['eslint']);
     grunt.registerTask('cs:server', 'Check PHP coding style', ['composer:run-script cs']);
--- a/package.json
+++ b/package.json
@@ -7,6 +7,8 @@
     "url": "https://github.com/SSilence/selfoss.git"
   },
   "devDependencies": {
+    "bootprint": "^1.0.2",
+    "bootprint-openapi": "^1.1.0",
     "grunt": "^0.4.5",
     "grunt-auto-install": "^0.3.1",
     "grunt-cli": "^1.2.0",

PR #1015, opened 2018-01-11
Add machine-readable API description

jtojnar:
This patch adds the description of the RESTful API in OpenAPI 3.0 format. Previously, the API was manually documented on the wiki without any guarantee that the description matches reality. Formal description will allow us to check for compliance automatically, as well as use other cool tools like Swagger UI.

For now, you can view the API on https://app.swaggerhub.com/apis/jtojnar/selfoss/2.18-SNAPSHOT

Closes: #993

cc @niol

### DIFF
--- a/docs/api-description.json
+++ b/docs/api-description.json
@@ -1,0 +1,1113 @@
+{
+  "openapi": "3.0.0",
+  "servers": [],
+  "info": {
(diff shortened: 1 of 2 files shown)


niol:
From a quick check:
- ```GET /logout``` seems to be missing
- ```GET /update``` seems to be missing
- ```GET``` and ```POST /opml``` seem to be missing, ```GET /opmlexport``` also
- ```GET /items/sync``` is missing
- the documented calls are looking consistent with the current implementation

jtojnar:
I only documented the “stable” API calls listed on the wiki. I think `/items/sync` is internal and only used by web client at the moment, `/logout` and `/update` should be documented though.

PR #1022, opened 2018-02-24
Add config option to hide the thumbnails

lucash:
The thumbnails, e.g. of the Youtube spout bloat the layout of the single
items while collapsed. This allows the user to hide the thumbnails via a
config option.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,9 @@
 # selfoss news
+## 2.19 – unreleased
+### New features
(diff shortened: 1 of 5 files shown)


jtojnar:
Hi @lucash. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contribution under `GPL-3-or-later`?

lucash:
Hi @jtojnar, I'm fine with licensing my contribution under `GPL-3-or-later`.

PR #1027, opened 2018-03-15
Fix pl translation

skotniczny:
pl.ini breaks selfoss 2.18. 

This pull request solves it. Translation also follows polish grammar rules, which was impossible [before](https://github.com/SSilence/selfoss/pull/556).


public/lang/pl.ini
```diff
@@ -1,13 +1,14 @@
...
 lang_newest=najnowsze
 lang_unread=nieprzeczytane
 lang_starred=z gwiazdką
 lang_tags=Tagi
 lang_alltags=wszystkie
 lang_timestamp={0,date} {0,time}
-lang_hours={0,plural,one {godzinę},two {# godziny},other {# godzin}} temu
-lang_minutes={0,plural,one {minutę},two {# minuty},other {# minut}} temu
-lang_seconds={0,plural,one {sekundę},two {# sekundy},other {# sekund}} temu
+lang_hours=przed {0,plural,one {godziną},other {# godzinami}}
```
jtojnar: I assume this is done to fix the incorrect “4 godzin temu”. Would you more expressive language files with [CLDR rules](https://www.unicode.org/cldr/charts/30/supplemental/language_plural_rules.html#pl) support (`{0,plural,one {godzinę},many {# godziny},other {# godzin}} temu`) be better or does this “przed 2 godzinami” sound good enough? It is something I want to do anyway, the current system is terrible, but I can prioritize it.

skotniczny: First of all, selfoss 2.18 with pl language gives error `TypeError: pluralValue is undefined` or `Uncaught TypeError: Cannot read property 'replace' of undefined`;  [there is no](https://github.com/SSilence/selfoss/blob/1246ce749cdfc963bf6d5f5b8e8ba1e9c9c3289c/public/js/selfoss-ui.js#L149) `two` case.

And yes, it fixes incorrect "3 godzin", "4 godzin", "23 minut", "24 minut", "33 sekund", … 

>Would you more expressive language files with CLDR rules support ({0,plural,one {godzinę},many {# godziny},other {# godzin}} temu) be better or does this “przed 2 godzinami” sound good enough?

Yes, indeed. This case is _not worth the effort_, "przed 2 godzinami"  probably sounds as good as "před 2 hodinami". :)

There is example how correct sentence with `… temu` should look:
 
```javascript
function format(num, unit) {
  let n1 = num % 10;
  let n2 = num % 100;
  let dictionary = {
    "sekundy": ["sekundę", "sekundy", "sekund"],
    "minuty": ["minutę", "minuty", "minut"],
    "godziny": ["godzinę", "godziny", "godzin"]
  };
  unit = dictionary[unit];
  if (n2 > 4 && n2 < 22) {
      return `${num} ${unit[2]} temu`;
  }
  if (n1 > 1 && n1 <= 4) {
      return `${num} ${unit[1]} temu'`;
  }
  if (num == 1) {
      return `${unit[0]} temu`; // number omitted intentionally
  }
  return `${num} ${unit[2]} temu`;
}

for (let i = 0; i <= 60; i++) {
  console.log(format(i, "sekundy"));
  console.log(format(i, "minuty"));
  console.log(format(i, "godziny"), "\n");
}
```

jtojnar: Right, we already have a client side implementation of pluralization rules, I agree that the effort is too high. I will add the CLDR support once we move everything to the client side.

jtojnar:
Thanks.

jtojnar:
Hi @skotniczny. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions under `GPL-3-or-later`?

skotniczny:
I'm fine with that.

Issue #1032, opened 2018-04-19
Reddit spout JSON parse problem


toonn:
It seems like the reddit spout has gotten me past the throttling problem but I think it expects a different JSON structure or something.
Spout [url](https://www.reddit.com/r/GameDeals/search?q=free+OR+100+nsfw%3Ayes&restrict_sr=on&sort=new&t=all). See #1024 for context.
```
[2018-03-09 09:17:59] selfoss.ERROR: error loading feed content for r/GameDeals {"exception":"[object] (GuzzleHttp\\Exception\\ParseException(code: 0): Unable to parse JSON data: JSON_ERROR_SYNTAX - Syntax error, malformed JSON at /srv/http/selfoss/vendor/guzzlehttp/guzzle/src/Message/Response.php:148)
[stacktrace]
#0 /srv/http/selfoss/spouts/reddit/reddit2.php(105): GuzzleHttp\\Message\\Response->json()
eddit2->load(Array)s/helpers/ContentLoader.php(99): spouts\
#2 /srv/http/selfoss/helpers/ContentLoader.php(53): helpers\\ContentLoader->fetch(Array)
#3 /srv/http/selfoss/controllers/Sources.php(290): helpers\\ContentLoader->updateSingle('53')
#4 /srv/http/selfoss/vendor/bcosca/fatfree-core/base.php(1806):controllers\\Sources->update(Object(Base), Array, 'controllers\\\\Sou...')
#5 /srv/http/selfoss/vendor/bcosca/fatfree-core/base.php(1627): Base->call(Array, Array, Array)
#6 /srv/http/selfoss/index.php(80): Base->run()
#7 {main}
"}
```

jtojnar:
I guess the issue is we are adding `.json` to the end of the URL, which results in `https://www.reddit.com/r/GameDeals/search?q=free+OR+100+nsfw%3Ayes&restrict_sr=on&sort=new&t=all.json`, instead of the `https://www.reddit.com/r/GameDeals/search.json?q=free+OR+100+nsfw%3Ayes&restrict_sr=on&sort=new&t=all` we want.

jtojnar:
Maybe try `r/GameDeals/search.json?q=free+OR+100+nsfw%3Ayes&restrict_sr=on&sort=new&t=all&foo=` for now.

toonn:
[...]

PR #1033, opened 2018-04-19
Reddit spout fixes

jtojnar:
Allow wider range of URLs in Reddit spout, including absolute URLs and searches.


toonn:
Fixed #1032 for me.

Issue #719, opened 2015-11-30
feat. req.: Share with anything on Android


aslmx:
Hi,

regularly i want to send the link to an feed item from the mobile selfoss via either text message or jabber/xmpp client.

Is there a simple way to integrate a configurable (share setting) button that will share the textline of the URL to the Android share handler (what ever it is called) so that it easily possible to share the URL with any application that handles text? 

would be a very great help and time saver!

Now i need to close the item/article, tap+hold the feed-items favicon and then copy the link, open whatever application and paste it. It also "destroys" the "workflow" when you use "foward ->" button...

aslmx:
just learnt that this seems to be a browser thing. While the Samsung default browser does not offer a share menu for links, android-firefox does and hence allows me to easily share the URL of a feed item.

I guess the feature request is then depending on a (non existing?) API between html/js and the browser.

can probably be closed if no one else has an idea how this could be easily implemented on selfoss side without the dependency to rely on a specific browser software on client side.

jtojnar:
I am not aware of any API for this.

jtojnar:
There is [Navigator.share](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/share) now, see https://github.com/SSilence/selfoss/pull/1017.

aslmx:
that sounds promising. can't wait for 2.19 :)

PR #1035, opened 2018-04-23
client: add native sharer

jtojnar:
Using  <a href="https://developer.mozilla.org/en-US/docs/Web/API/Navigator/share">Web Share API</a> for browsers that support it

Closes: #719


jtojnar:
It is not supported outside of Chrome, and even there it is behind a flag. It probably should not be merged yet.

jtojnar:
This is [supported](https://caniuse.com/#feat=web-share) by Safari in addition to Chrome on Android so it is time to add this.

jtojnar:
I successfully shared a link on Chrome for Android. Though, currently, the sharers have a bug, where they will include some accessibility markup.

Also on desktop Chromium, the button is shown, even when there are no targets, resulting in `DomException`. Looking at the [spec](https://w3c.github.io/web-share/#share-method), it does not appear to be possible to distinguish the case when there are no sharers (6.1) from one where user cancels the sharing (6.3)

jtojnar:
Also this feature is secure-context only (HTTPS).

PR #1052, opened 2018-09-20
Use Guzzle with SimplePIe

jtojnar:
SimplePie uses custom code for fetching resources from web but this is inconsistent and error-prone. To fix this, we will be overloading SimplePie’s `File` class with our own subclass using Guzzle.

Additionally, we will be no longer adding “Simplepie/X.Y.Z” string to the `User-Agent` header. Given that this is just an implementation detail, and that many websites actively block SimplePie.

### DIFF
--- a/helpers/SimplePieFileGuzzle.php
+++ b/helpers/SimplePieFileGuzzle.php
@@ -1,0 +1,47 @@
+<?php
+
+namespace helpers;
+
+/**
+ * Bridge to make SimplePie fetch resources using Guzzle library
+ */
+class SimplePieFileGuzzle extends \SimplePie_File {
+    public function __construct($url, $timeout = 10, $redirects = 5, $headers = [], $useragent = null, $force_fsockopen = false) {
+        $this->url = $url;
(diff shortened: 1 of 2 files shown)


akash07k:
Can't we just expose the user agent as: Mozilla/5.0 (Windows NT 10.0; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/77.0.3865.75 Safari/537.36
Because it is the standard user agent for chrome and none of the websites will block it. and, I'm wondering, that why the autodiscovery of the feeds is broken due to this?

jtojnar:
Just for completeness, the user agent change did not cause this, the other change did.

And as for the user agent change, we already are somewhat Chromey. If we should pretend to be Chrome is a separate issue.

akash07k:
> Just for completeness, the user agent change did not cause this, the other change did.
> 
> And as for the user agent change, we already are somewhat Chromey. If we should pretend to be Chrome completely is a separate issue.

Oh, ok, :-)

PR #1057, opened 2018-10-08
Handle feeds with date that cannot be parsed

JFreakDK:
I have a feed where SimplePie cannot parse the pubDate. The format looks like this: "<pubDate>ma, 08 okt 2018 10:09:29 +0000</pubDate>".
The result is a date for the item that is epoch (which is too old). Therefore the items will never show up in the reader as they are too old. By falling back to current time when the date cannot be parsed will make the items show up in selfoss.


jtojnar:
Do not forget to report the wrong date formatting to the website itself.

PR #1058, opened 2018-10-09
return empty array instead of nil tag

### LINKED ISSUE #1056: Sources without tags return nil tag in API
On master, I [noticed](https://github.com/SSilence/selfoss/issues/989#issuecomment-427603398] that the `/items` API endpoint returns an empty tag name for items whose sources do not have any tags (the tags field in the sqlite database is an empty string).

```json
    …
    "tags": [
      ""
    ]
    …
```

niol:
This would fix #1056 .


daos/mysql/Statements.php
```diff
@@ -163,7 +163,11 @@ public function ensureRowTypes(array $rows, array $expectedRowTypes) {
                             }
                             break;
                         case \daos\PARAM_CSV:
-                            $value = explode(',', $row[$columnIndex]);
+                            if ($row[$columnIndex]) {
```
jtojnar: Maybe use explicit ` === ''` check? This would probably filter out `0` tag.

Issue #979, opened 2017-08-17
Consolidate the documentation


jtojnar:
At the moment documentation is spread between the README, website and wiki, often with information duplicated. We should move the documentation to the repository and generate the website from it and README.

akash07k:
Really a great idea of consolidating the documentation indeed.
I'm wondering that why it got so delayed. :(

jtojnar:
Like many other open-source software project, selfoss does not have many developers. And when we find some time to spare, we tend to spend it on fixing bugs or disentangling the whole jQuery spaghetti.

This is also made complicated by the fact that we need to redesign the website to accommodate for all these new pages.

akash07k:
Ya bro, you are right.
I'll try my best to write the documentation if you want.

jtojnar:
The hardest part is not even writing the docs but coming up with a website structure.

I classified the sections on the current website based on how much it makes sense to me to include them:

> Introduce selfoss, okay:
> 
> * features
> * screenshots
> 
> Show how to set it up, not sure:
> 
> * requirements
> * installing
> * configuring
> * importing
>
> Show advanced usage, less sure:
> * shortcuts
> * updating
> * list of configuration options – too long
> 
> For developers, probably not:
> 
> * Extending selfoss
> * API
> 
> Credits and legal info, maybe still:
> 
> * License
> * About

I based this mostly on the assumption that the main page should attract new users. It should not contain unnecessary details but it also should not be just huge scrolling wallpaper with no content (see [Wallabag](https://www.wallabag.org/en) website).

Looking at websites of few other self-hosted apps for inspiration:

* https://miniflux.app/ – website not very shiny but [documentation](https://miniflux.app/docs/index.html) is divided into admin, user & developer sections
* https://tt-rss.org/ – not well structured, just a landing page + [unstructured wiki](https://tt-rss.org/wiki.php)
* https://freshrss.org/ – very nice landing page, content-wise close to something I had in mind; [docs](https://freshrss.github.io/FreshRSS/en/) also split into admin, user & developer sections but navigation there is not very clear
* https://www.wallabag.org/en ­– main page lacks info, docs link is at the bottom, plus it brings scrollmageddon; [docs](https://doc.wallabag.org/en/) in GitBook again with the same split
* https://vikunja.io/ – I love this site, especially the [docs](https://vikunja.io/docs/) section is well structured, though maybe I would merge home and features
* https://matomo.org/ – looks too business oriented, with case studies

PR #1060, opened 2018-10-20
Improve Readme & default.ini

Binnette:
Hi @jtojnar, here are some ideas in order to improve Readme and default.ini.
Feel free to accept/comment/reject them. 🌞

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,9 +18,6 @@ INSTALLATION
 
 1. Upload all files of this folder (IMPORTANT: also upload the invisible .htaccess files)
 2. Make the directories data/cache, data/favicons, data/logs, data/thumbnails, data/sqlite and public/ writeable
-
-   `chmod o+w selfoss/data/cache selfoss/data/favicons selfoss/data/logs selfoss/data/thumbnails selfoss/data/sqlite selfoss/public/`
-
 3. Insert database access data in config.ini (see below -- you don't have to change anything if you want to use sqlite)
 3. You don't have to install the database, it will be created automatically (ensure that your database has enought rights for creating triggers)
 4. Create cronjob for updating feeds and point it to https://yourselfossurl.com/update via wget or curl. You can also execute the cliupdate.php from commandline.
@@ -69,7 +66,7 @@ OPML Import
 Selfoss supports importing OPML files. Find the OPML export in the old application, it is usually located somewhere in settings. Then visit the page https://yourselfossurl.com/opml and upload it there.
 
 
-MOBILE APPS
+THIRD-PARTY APPS
 ----
 
 | App Url            | OS      | Project status | Compatibility* |
--- a/README.md
+++ b/README.md
@@ -18,6 +18,9 @@ INSTALLATION
 
 1. Upload all files of this folder (IMPORTANT: also upload the invisible .htaccess files)
 2. Make the directories data/cache, data/favicons, data/logs, data/thumbnails, data/sqlite and public/ writeable
+
+   `chmod o+w selfoss/data/cache selfoss/data/favicons selfoss/data/logs selfoss/data/thumbnails selfoss/data/sqlite selfoss/public/`
+
 3. Insert database access data in config.ini (see below -- you don't have to change anything if you want to use sqlite)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -66,10 +69,15 @@ OPML Import
...
 
-APPS
+MOBILE APPS
 ----
 
-Two third party apps are available for Android: [Selfoss](https://play.google.com/store/apps/details?id=fr.ydelouis.selfoss) and [Reader For Selfoss](https://play.google.com/store/apps/details?id=apps.amine.bou.readerforselfoss).
+| App Url            | OS      | Project status | Compatibility* |
+|--------------------|---------|----------------|----------------|
+| [Reader For Selfoss](https://play.google.com/store/apps/details?id=apps.amine.bou.readerforselfoss) | Android | Active         | 2.18 and older |
+| [Selfoss](https://play.google.com/store/apps/details?id=fr.ydelouis.selfoss)            | Android | Inactive       | 2.18 and older |
```
jtojnar: I would just remove the inactive app from the list.

README.md
```diff
@@ -66,10 +69,15 @@ OPML Import
 Selfoss supports importing OPML files. Find the OPML export in the old application, it is usually located somewhere in settings. Then visit the page https://yourselfossurl.com/opml and upload it there.
 
 
-APPS
+MOBILE APPS
```
jtojnar: Maybe this could be called third-party apps, since I want to add support to https://github.com/jangernert/FeedReader once we stabilize the API for 3.0.

jtojnar: Though, I am not sure if we want to include this section in the README. I am still struggling with what should go where: README, website or both.

README.md
```diff
@@ -18,6 +18,9 @@ INSTALLATION
 
 1. Upload all files of this folder (IMPORTANT: also upload the invisible .htaccess files)
 2. Make the directories data/cache, data/favicons, data/logs, data/thumbnails, data/sqlite and public/ writeable
+
+   `chmod o+w selfoss/data/cache selfoss/data/favicons selfoss/data/logs selfoss/data/thumbnails selfoss/data/sqlite selfoss/public/`
```
jtojnar: Many users will be uploading selfoss to shared hosts where they do not have terminal access. I think I would rather keep the readme simple and explain it in detail in a separate installation guide.

defaults.ini
```diff
@@ -10,6 +10,7 @@ db_password=
 db_port=
 db_prefix=
 logger_destination=file:data/logs/default.log
+; logger_level values: EMERGENCY, ALERT, CRITICAL, ERROR, WARNING, NOTICE, INFO, DEBUG, NONE
```
jtojnar: I am not sure we want to duplicate the docs in the config. Instead I would update the link at the top to point to https://selfoss.aditu.de/#configuration_params, add [config validation](https://github.com/SSilence/selfoss/issues/1061) and [streamline the docs](https://github.com/SSilence/selfoss/issues/979).

README.md
Binnette: No, i use mobile app 'Selfoss' everyday, Ok the project is inactive but still compatible with Selfoss. I will use this this mobile app unless "Reader For Selfoss" add an "offline mode".

jtojnar: Unfortunately, it does not work with 2.19 so we will have to remove it anyway. Hopefully, by the time it is released, Reader for Selfoss will support offline mode: https://github.com/aminecmi/ReaderforSelfoss/issues/179

defaults.ini
```diff
@@ -1,5 +1,5 @@
-; see https://selfoss.aditu.de for more information about
-; the configuration parameters
+; see https://selfoss.aditu.de/#configuration_params
+; for more information about he configuration parameters
```
jtojnar: You dropped “t” in “the”

Binnette: Oooops. Fixed :wink:

README.md
aminecmi: @Binnette There is now an offline mode.

jtojnar:
Fixed in 4383a63b. Thanks.

PR #1064, opened 2018-10-21
Render sources in navigation client-side

jtojnar:
cc @niol


jtojnar:
In the future, we can replace this with virtual-dom-based templates (e.g Vue.js); for now, this should not be much worse than the current solution.

akash07k:
Is this PR still in progress? or we can merge it now?

jtojnar:
Yeah, this is WIP. I will revisit it after #1096

akash07k:
Ok, great.

jtojnar:
Surpassed #1184

PR #1054, opened 2018-10-03
Front end i18n updates

Seanmcn:
Resolves #1000 adding in the suggested translation in French, the rest with Google Translate..

Adds variables in a number of other places on the front-end where there is hardcoded English. 

Happy #hacktoberfest 😄

PR #1014, opened 2018-01-11
Make selfoss an offline web application

niol:
Here are a couple of patches that make selfoss an offline web application. I've been using this for some months now and fixed the last issues I could encounter a few weeks ago. This has been working quite well for me in planes or subways with no network access and the app seamlessly syncs back to the server when available.
This pull request contains a lot of code but it is mostly different code paths from the current online behavior, that's why regression are unlikely. I'd be happy to fix any issue reported with this.

PR #1073, opened 2018-12-11
Add Css and  Image To Buttons(Add Import Export)

hamedhossani:
It's easier to work with css style and icons.Without them I could hardly work.


jtojnar:
What do you mean it to be easier to work with?

controllers/Sources.php
```diff
@@ -25,7 +25,7 @@ public function show() {
 
         // load sources
         $sourcesDao = new \daos\Sources();
-        echo '<button class="source-add">' . \F3::get('lang_source_add') . '</button>' .
+        echo '<button class="source-add">' . \F3::get('lang_source_add') . '</button>' .' or '.
```
jtojnar: Having or in the buttons is not very nice but adding it here is even worse, since it cannot be translated.

public/css/style.css
```diff
@@ -842,12 +844,33 @@ body * {
     
 }
 
-.source-export,
+.source-export{
+    font-size:0.60em;
+    text-decoration:none;
+    color:#777;
+    margin-left:10px;
+    background: url(images/export.png) no-repeat left center #65a06c;
```
jtojnar: I am not really sure I want to add icons to imports and exports, since icons bring attention to things and I feel these action are used too rarely to warrant that.

public/css/style.css
```diff
@@ -842,12 +844,33 @@ body * {
...
+    height: 30px !important;
+    /* line-height: 30px; */
+    display: inline-block;
+}
 .source-opml {
     font-size:0.60em;
     text-decoration:none;
     color:#777;
     margin-left:10px;
+    background: url(images/import.png) no-repeat left center #65a06c;
```
jtojnar: Moreover we are working on rewriting the icons in #1013. It is just next in queue after #1014 and #1054.

jtojnar:
Closing since there has been no activity.

Issue #1021, opened 2018-02-09
Infinite scroll, good ! Mark all as read, bad !


KewlCat:
I like the infinite scroll, but being interrupted isn't. I'd like to have a way to "mark as read" until a specific element (this, or automatically mark as read anything that's been scrolled away from the top ?) because when I'm interrupted there are many elements on the screen that I've read but not marked, one element that I was reading, and many other elements I haven't even seen on screen. Marking all of them as read is too violent (I'll be missing the ones I had not read) and leaving it as-is before a refresh means I'll re-read the ones I haven't marked as read.
I have no idea for an implementation of this. Do you ?

aslmx:
You're doing it wrong! SCNR 😁

I suggest you read your feeds by tag or by source. This way you can work through your items step by step. 
This is how i do it since 4 years or so. If i i want to read something later which would bei marked as read, i star it.

HTH!

KewlCat:
How dare you pretend I'm using it wrong !? I've been using it for... what... 24hours ? :-P
Seriously though, I've ported all my (bad?) habits from the previous reader and I'm trying to find my new workflow.
Currently it's one of : 
 - not interested (Mark as read and foget)
 - I'll read that right after I'm finished (open in a new tab)
 - I'm not working on this right now but it'll be useful one day, maybe (Wallabag)
I don't have too many sources, but some are very prolific, so I try to skim as fast as possible (hence scrolling and having them marked automatically is great for me)
I'll try using it "by tags".
As for starring, I know I don't come back to read them later (I had hundreds of starred articles in Google Reader when it shut down, I never had the chance to read them fully). So now either I read them right after I "finish" with the new items in my RSS reader, or I add them to Wallabag and when I search for the specific topic they cover, they come back up and I can read them. They are read "when needed".

That helped, thank you. Hope my explanation helps make some sense of all my remarks.

niol:
For this, I disabled ```auto_stream_more``` which makes the items paginated, and then I mark as read page after page.
Your request may be the same as #681 .

PR #1086, opened 2019-02-03
added 'read up to here' button on entries in articleList, moved open utton to top of toolbar (was to close to readuptohere)

otdftr:
a feature I found very useful in NEWSIFY: mark read "up to here" 
hacked into my v2.18 installation of Selfoss and wanted to share it as a suggested feature


niol:
This is a good idea. This may be an answer to #1021 .
As for the code, I would suggest not including history comments in the source and less code duplication (creating `selfoss.dbOnline.mark(ids)` and `selfoss.ui.mark(ids)` may help though handling the possible rollback may be tricky). Hopefully I can fix that later if this get accepted.

otdftr:
Feel free to strip comments etc., I included them for peer review only.

jtojnar:
Sorry, forgot to reply here.

I like this idea and occasionally wanted this myself but I wanted to avoid cluttering the interface with more buttons so until we come up with a way to let users customize the buttons (perhaps extend the `share` config to other buttons as well) I would suggest users to add the snippet to their `user.js` file and am concerned about the code duplication.

I think `selfoss.dbOnline.mark(ids)` would be very useful since the `markVisibleRead` is already beyond tolerable complexity.

Issue #1089, opened 2019-02-10
Cannot view all items


tuur29:
#1014 added the functionality to work offline and it's great. 

However I can no longer access my old read items. I can see the amount of items next to the newest menu item (16/1641) and see those 16 items loaded after the upgrade, more are downloaded when I refresh my sources. But I cannot see all (non-offline) items. They do show when searching.

This problem might be limited to my fork, but the intended functionality isn't clear to me which makes this very hard to debug. Should the application be downloading them in the background or load them when I view /newest/all or am I missing something else?

niol:
As a workaround, you can always disable offline by logging out.
Regarding items that are not in the offline db, selfoss should know when items are to be found online only. It should switch from querying the offline db to requesting `/?offset...` on the server automatically. 
The intent is to have unread and starred items available offline, and then cleanup items older by 10 days, not unread, not starred. Maybe there's a bug somewhere, but I've tested this numerous times.

When you say that you cannot access your old read items, is it that the "more" button does not show?

tuur29:
Aah then it's working as intended. It's only syncing items from the last 10 days + unread and starred. Is there a reason why you don't just sync all items?

I can access them in normal mode just fine, thanks.

niol:
Syncing all items is too much for offline storage (especially on mobile devices), given that users may have a large number of items.
Regarding accessing old read items, let me repeat myself: you should not have to log out and log in again without offline storage to read them. If that's the case, there's a bug I need to look into: selfoss should switch from querying the offline db to requesting on the server automatically.

tuur29:
Alright.
I looked around for a possible cause and found my problem happens after reloading the page with devtools opened and the "disable cache" flag enabled on the network tab. Once I do this, the only way to fix it is to login again.
I've tested this on master and the problem pops up there as well.

niol:
Opening devtools and disabling cache should not modify the behaviour. When you say that you cannot access your old read items, is it that the "more" button does not show?

tuur29:
That is correct. I have `auto_stream_more` enabled with `items_perpage=50` with 40 items synced. Changing either of these settings have no effect on the problem. After loading 40 items the more buttons doesn't show.

niol:
I'll look into this. Maybe `auto_stream_more` enabled or `items_perpage > ` items sync is causing problems.

niol:
OK this was a problem with `items_perpage > ` items synced, the following should fix this:

```patch
[...]

PR #1090, opened 2019-02-17
fix all items when offline synced items less than items per page (#1089)

niol:


jtojnar:
Can you ask Tobias for commit access? I will not have access to my computer soon and merging from the Web UI does not make for very pretty commit history.

niol:
I'v asked Tobias but I'm not sure what "very pretty commit history" means in that case.

jtojnar:
I mean trying to maintain a linear history using `rebase` command compared to the convoluted graph created by `git merge`.

niol:
I think that as the parent of this pull request commit is the last commit of master, `git merge` won't create a merge commit et rebasing is not needed.

jtojnar:
Yeah, in this case `git merge` would fast forward but the GitHub UI explicitly does not do fast forward merges.

SSilence:
niol I invited you as collaborator

PR #1091, opened 2019-02-17
new lastread filter

niol:
Introduces a new filter: last read items, that is items ordered by the last time they were read, excluding the items that were skipped using 'Mark Visible Read'.

### DIFF
--- a/controllers/Items.php
+++ b/controllers/Items.php
@@ -30,14 +30,17 @@ class Items extends BaseController {
             $lastid = $_POST['ids'];
         }
 
+        $skipped = array_key_exists('skipped', $_POST) && $_POST['skipped'] == 'true';
+
         $itemDao = new \daos\Items();
 
         // validate id or ids
         if (!$itemDao->isValid('id', $lastid)) {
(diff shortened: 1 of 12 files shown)


jtojnar:
Do we want filters to change the order of items? And will it not be confusing that marking items individually as read behaves differently from using the green button?

niol:
I've found it very useful for me to have some kind of reading history distinguishing between items I opened and items I just marked as read without reading. This makes it possible to quickly find an article the you read a couple hours before, something that is more difficult without this.
Also, this opens the door to having stats of how much article you read regarding how much you discard (not implemented here).

jtojnar:
One thing I do not like on this is that the `skipped` status is a separate column, even though it is mutually exclusive with the `unread` flag. Maybe it would make more sense to replace both columns with single `status` column that would be of type `enum(new, read, dismissed)`.

jtojnar:
Or even simpler, we could track `readtime` and `readtime == null && unread == 0` would mean the item was dismissed (though this is maybe too implicit).

PR #1093, opened 2019-02-20
Get rid of Grunt

jtojnar:
A prerequisite to Webpack. (I could not get grunt-webpack task to work.)

Still need to check the zipball is created correctly.


jtojnar:
Makefile might be cleaner solution but I am not sure how available make is on Windows.

jtojnar:
Okay, the zipball now contains all files the original one did. We also have some extra FTR files starting with dot like `vendor/j0k3r/graby-site-config/.wordpress.com.txt` which were previously filtered out.

PR #1096, opened 2019-03-03
Port to Webpack

jtojnar:
Since I did not have time to finish this this weekend either, I have decided to push it here as is for now.

It is mostly working, with the exception of serviceworker – we need some reliable webpack plug-in for generating the list of files the service worker should cache and also for invalidating the worker.

I also want to generate index.html so that we do not have to care about it in PHP. That will require API for obtaining config values.


jtojnar:
Closing in favour of #1137

PR #1098, opened 2019-03-23
helpers/SimplePieFileGuzzle: handle case where headers is given but is null.

desaintmartin:


helpers/SimplePieFileGuzzle.php
```diff
@@ -6,10 +6,13 @@
  * Bridge to make SimplePie fetch resources using Guzzle library
  */
 class SimplePieFileGuzzle extends \SimplePie_File {
-    public function __construct($url, $timeout = 10, $redirects = 5, $headers = [], $useragent = null, $force_fsockopen = false) {
+    public function __construct($url, $timeout = 10, $redirects = 5, $headers = null, $useragent = null, $force_fsockopen = false) {
         $this->url = $url;
         $this->permanent_url = $url;
         $this->useragent = $useragent;
+        if ($headers == null) {
```
jtojnar:
```suggestion
        if ($headers === null) {
```

jtojnar:
For reference, this is where it is called with `null`:

https://github.com/simplepie/simplepie/blob/11ddaed291ccd1b2628eef433a6bef58a1e4f920/library/SimplePie/Locator.php#L411

jtojnar:
Thanks for fixing this.

jtojnar:
Hi @desaintmartin. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contribution under `GPL-3-or-later`?

desaintmartin:
As someone who wrote gpl code for years for a living, I could not be more happy if selfoss is GPL-3-or-later! :)

PR #1102, opened 2019-04-03
spouts: Port Twitter spouts to Guzzle

jtojnar:
This concludes our effort to port selfoss to Guzzle.

https://www.sitepoint.com/using-guzzle-twitter-via-oauth/


jtojnar:
Extensive testing did not reveal any issues so merging.

PR #1124, opened 2019-08-08
Fix #1123 Google news alerts

### LINKED ISSUE #1123: No icon on google news alert feed
After adding rss feed
https://www.google.de/alerts/feeds/18120933104492874959/12861401000818240705
i got no favicon.

Themroc:
Seems to be a simplepie bug, it doesn't give back URL from < link href="URL" rel="self"/ >
I'd rather not touch that, so here's a workaround.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -1130,53 +1130,6 @@
             "time": "2019-09-06T13:49:17+00:00"
         },
         {
-            "name": "natxet/CssMin",
-            "version": "v3.0.6",
-            "source": {
-                "type": "git",
-                "url": "https://github.com/natxet/CssMin.git",
-                "reference": "d5d9f4c3e5cedb1ae96a95a21731f8790e38f1dd"
-            },
-            "dist": {
-                "type": "zip",
-                "url": "https://api.github.com/repos/natxet/CssMin/zipball/d5d9f4c3e5cedb1ae96a95a21731f8790e38f1dd",
-                "reference": "d5d9f4c3e5cedb1ae96a95a21731f8790e38f1dd",
-                "shasum": ""
-            },
-            "require": {
-                "php": ">=5.0"
-            },
-            "type": "library",
-            "extra": {
-                "branch-alias": {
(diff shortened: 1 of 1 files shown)


jtojnar:
Sorry, it took so long to get back to you.

I have replaced 0b5c57cb630f554bdd375d3963ee242087f45397 with “spouts/rss/feed: fallback to feed domain icon”. That should be more correct, since the feed’s `self` URL is not `htmlUrl` (that is reserved for the HTML equivalent of the feed, for example category listing, or user’s Twitter timeline).

Actually, `params['url']` can contain address of the website when autodiscovery is used. In that case, it would be correct, but we do not know if it is the case.

Unfortunately, it might also be not enough in some cases, since `subscribe_url()` does not return the effective URL, even though the [documentation](http://simplepie.org/wiki/reference/simplepie/subscribe_url) says it should. It is a bug in SimplePie: https://github.com/simplepie/simplepie/issues/2#issuecomment-537958256

f4f70a5fe9c50244046d6274632534c792089377 was removed in favour of “helpers/webclient: track redirects”. Passing data through object state is brittle and ugly.

30cd7a077711b4c815e411b67b5dd3c0be840440 took care of multiple things: returning the support for uppercase, improving the icon selection algorithm, and fixing the handling of relative paths. I have split it into two commits and improved coding style to be more in line with selfoss.

jtojnar:
And sorry for accidentally pushing wrong head. Here is the updated branch: https://github.com/SSilence/selfoss/tree/1123

jtojnar:
I opened a PR against SimplePie fixing the other side of the issue: https://github.com/simplepie/simplepie/pull/627

jtojnar:
Hi @Themroc. Previously, it [was not clear](https://github.com/fossar/selfoss/issues/1218) if selfoss was licensed under GPL 3 only, or also any later version. Could you clarify whether you are fine with licensing your contributions here and in other PRs under `GPL-3-or-later`?

Issue #1130, opened 2019-09-15
Feed autodiscovery broken


akash07k:
I've just installed SelfOss from latest master branch and configured everything.
But when I try But when I try to import existing opml / try to add a new feed, I get some errors like "xml contains invalid characters" etc.
What can be the issue?
and, when can we expect 2.19 release?

jtojnar:
Do you think you could share the OPML file? You can use my private mail jtojnar@gmail.com if you don't want to share publicly. Or you could try to check the feed with https://validator.w3.org/feed/.

As for the new release, I intended to finish it during the summer but other work got in the way. The main blocker is the script caching in the offline mode, which I am not completely happy about.

akash07k:
@jtojnar jtojnar
I've attached the xml.
Also, checked with the w3validator, no issues found.
Please check.
Changed the extention because .xml files are not allowed.
[selfoss-subscriptions (1).txt](https://github.com/SSilence/selfoss/files/3614077/selfoss-subscriptions.1.txt)

akash07k:
@jtojnar 
But I really like selfoss.
It has changed my life completely.
you guys have developed a very excellent piece of software.

jtojnar:
What feeds specifically do fail for you?

akash07k:
All the feeds, except rss-ndtv

jtojnar:
Hmm, AndroidAuthority does not seem to declare the feed anywhere. You will need to set the URL manually: https://www.androidauthority.com/feed/

But for the other sites, it should work. For some reason the SimplePie autodiscovery does not work on them.

akash07k:
Hmm, I agree. but then why is it working properly in 2.18? this OPML is exported by selfoss 2.18 version itself.

jtojnar:
I would guess this might be the culprit: https://github.com/SSilence/selfoss/pull/1052

Or less likely, the SimplePie. I will try to take a look later. For now, you can use the direct link to the feed.

akash07k:
But it's not possible to find the direct link to the feed for every
website. Even, there are many websites for which miniflux and other
applications can't fetch feeds, but selfoss is able to do so. But due
to this issue in this latest branch, it's becoming difficult.

jtojnar:
[...]

PR #1131, opened 2019-09-16
SimplePieFileGuzzle: Fix autodiscovery

jtojnar:
In https://github.com/SSilence/selfoss/pull/1052, we introduced a class that allowed us to make SimplePie fetch feeds using Guzzle and then pass headers and content back to it. Unfortunately, SimplePie expects the headers to be lower-case and strings, whereas Guzzle returns original case and string arrays as mandated by PSR-7. This in turn broke autodiscovery.

Closes: https://github.com/SSilence/selfoss/issues/1130


jtojnar:
Regression in development version, no need to include it in changelog.

PR #1135, opened 2019-09-23
a11y: entry stream pagination

niol:


public/js/selfoss-ui.js
```diff
@@ -138,6 +138,21 @@ selfoss.ui = {
...
+                for (var pageNumber = 1;
+                    pageNumber < count / selfoss.filter.itemsPerPage + 1;
+                    pageNumber++) {
+                    var currentAttr = '';
+                    var current = '';
+                    if (selfoss.filter.offset == selfoss.filter.itemsPerPage * (pageNumber - 1)) {
+                        currentAttr = ' aria-current="true"';
+                        current = ' (current)';
+                    }
+                    streamPagination.append('<li><button aria-label="Goto Page ' + pageNumber + '"' + currentAttr + '>' + pageNumber + current + '</button></li>');
```
jtojnar: Could you make this localizable? And also `Go to page {0}`.

templates/home.phtml
```diff
@@ -204,6 +204,9 @@
             <button class="stream-button stream-more"><span><?= trim(\F3::get('lang_more')); ?></span></button>
             <button class="stream-button mark-these-read"><span><?= trim(\F3::get('lang_markread')); ?></span></button>
             <button class="stream-button stream-error"><?= trim(\F3::get('lang_streamerror')); ?></button>
+            <nav role="navigation" aria-label="Pagination Navigation">
+                <ul id="stream-pagination" />
```
jtojnar: This is not a XML document, so we should probably use closing tag.

public/js/selfoss-events-entries.js
```diff
@@ -169,6 +169,13 @@ selfoss.events.entries = function() {
         });
     }
 
+    // stream pagination buttons
+    $('#stream-pagination button').unbind('click').click(function() {
+        var pageNumber = parseInt($(this).html());
```
jtojnar: In this case, we control the values parsed but it would still be prudent to explicitly specify the radix.

```suggestion
        var pageNumber = parseInt($(this).html(), 10);
```

I am surprised eslint does not shout about this, [the rule](https://eslint.org/docs/rules/radix) should be on by default.

public/js/selfoss-db.js
```diff
@@ -559,19 +559,29 @@ selfoss.dbOffline = {
...
                         return true;
                     }
 
                     if (selfoss.filter.type == 'starred') {
-                        return entry.starred;
+                        keepEntry = entry.starred;
                     } else if (selfoss.filter.type == 'unread') {
-                        return entry.unread;
+                        keepEntry = entry.unread;
+                    } else if (selfoss.filter.type == 'lastread') {
```
jtojnar: This looks like a leak from https://github.com/SSilence/selfoss/pull/1091.

public/js/selfoss-ui.js
```diff
@@ -138,6 +138,21 @@ selfoss.ui = {
...
+                var count = parseInt($('.nav-filter-' + selfoss.filter.type +
+                                       ' span.count').html());
+                for (var pageNumber = 1;
+                    pageNumber < count / selfoss.filter.itemsPerPage + 1;
+                    pageNumber++) {
+                    var currentAttr = '';
+                    var current = '';
+                    if (selfoss.filter.offset == selfoss.filter.itemsPerPage * (pageNumber - 1)) {
+                        currentAttr = ' aria-current="true"';
+                        current = ' (current)';
```
jtojnar: This should be localizable as well as `Go to page {0} (current)`.

jtojnar:
I wonder if using offsets is a good idea with streams. Would not it be safer to use the seek pagination introduced in https://github.com/SSilence/selfoss/pull/869?

akash07k:
Implementing the pagination is really a wonderful idea for accessibility as well as usability benefits.
I'm also happy that you used aria-current as well.
Thanks a lot for this.

niol:
I'll take the comments into account later.
What's also missing is navigation hash updates (i.e. https://selfoss.example.com/#unread/all/page-3).

akash07k:
> I'll take the comments into account later.
> What's also missing is navigation hash updates (i.e. https://selfoss.example.com/#unread/all/page-3).

So, you will also work on that?

jtojnar:
@akash07k Would it work for you if we showed just _previous_ and _next_ buttons? Are you interested in pagination simply to limit the amount of content on the screen for easier navigation with screen reader, or do you need to be able to switch to arbitrary page?

@niol Could we make this configurable? I actually prefer to be able to scroll back to the previous items. Also, how does this interact with `auto_stream_more` option?

jtojnar:
> But if it is very difficult to achieve and implement, then we can just compromise with 'next' and 'previous' buttons only.

It should be possible to combine seek pagination with explicit pages (via seek+offset) but the question is if we should. Selfoss strives to be a simple program to use so we need to always ask if the feature would help someone achieve a concrete existing goal. Hypothetical use cases or the fact that other software provides the feature is not sufficient reason to implement it.

I cannot come up with a task that would jumping to a specific page achieve, so I want to know more about your use case. For example, for the “Next” button, we could say: “I have read all the news on this page and want to read more.”

Also if we wanted to keep the page numbers, we would need to trim the navigation somewhat to avoid too many buttons:

Screenshot of the items view in selfoss with this patch applied, scrolled to the bottom of the page, the 289 buttons for switching page occupying most of the screen

Having more information about your use case would allow us to make the pagination usable for you, while omitting everything that you do not need. I assume having screen reader to navigagate 289 buttons will not be pleasant either.

akash07k:
> > But if it is very difficult to achieve and implement, then we can just compromise with 'next' and 'previous' buttons only.
> 
> It should be possible to combine seek pagination with explicit pages (via seek+offset) but the question is if we should. Selfoss strives to be a simple program to use so we need to always ask if the feature would help someone achieve a concrete existing goal. Hypothetical use cases or the fact that other software provides the feature is not sufficient reason to implement it.
> 
> I cannot come up with a task that would jumping to a specific page achieve, so I want to know more about your use case. For example, for the “Next” button, we could say: “I have read all the news on this page and want to read more.”
> 
> Also if we wanted to keep the page numbers, we would need to trim the navigation somewhat to avoid too many buttons:
> 
> Screenshot of the items view in selfoss with this patch applied, scrolled to the bottom of the page, the 289 buttons for switching page occupying most of the screen
> 
> Having more information about your use case would allow us to make the pagination usable for you, while omitting everything that you do not need. I assume having screen reader to navigagate 289 buttons will not be pleasant either.

Actually, if it is too dificult then we can only have just previous and next buttons only.
Actually, we can only have 4-5 pagination links on the page and then we can have the links for first page and last page entirely along with the next and previous page links.
Actually, it can be needed in the cases, for example let's say 50 items are displaying on a page and we directly want to read the news from the past let's say from 4th page, then in this case it will help.
But again, if it is difficult and complicated, then we can only have just next and previous buttons/links.

niol:
There is no technical difficulty but I won't use this so cannot input on the use case. I'll follow-up on the navigation hash and configurability when what's wanted is more clear.

akash07k:
> There is no technical difficulty but I won't use this so cannot input on the use case. I'll follow-up on the navigation hash and configurability when what's wanted is more clear.

So, what will be the current implementation of the paginations as of now?

niol:
What's proposed here is adding page 1 to 5 buttons with current page indication. It works but needs ironing out.

akash07k:
> What's proposed here is adding page 1 to 5 buttons with current page indication. It works but needs ironing out.

Thanks, it's good. also, it will also be good if there will be pagination for the first and last page as well. BTW, thanks a tons for your effort bro. Till when can you complete it fully?

akash07k:
Is it ready for merge?

niol:
> 
> 
> I wonder if using offsets is a good idea with streams. Would not it be safer to use the seek pagination introduced in #869?

In order to seek, you need to know the last item of each page to be able to jump. offset is the only way.

niol:
> 
> 
> Is it ready for merge?

No. Missing:
- navigation hash updates
- see of this can be configurable
- add links to first and last pages
- rebasing

akash07k:
Oh, ok.
:(

akash07k:
> Oh, ok. :(
> […](#)
> On 10/3/19, niol ***@***.***> wrote: > > > Is it ready for merge? No. Missing: - navigation hash updates - see of this can be configurable - add links to first and last pages - rebasing -- You are receiving this because you were mentioned. Reply to this email directly or view it on GitHub: [#1135 (comment)](https://github.com/SSilence/selfoss/pull/1135#issuecomment-537935283)

@jtojnar 
@niol Guys, can you please complete this implementation soon?
We need it because infinite scrolling cause lots of problems and after a lot of entries gets loaded the page starts to hang due to the heaviness.
I'm just requesting, please don't feel bad, this is just asimple request. :(

niol:
I'm still not satisfied with what I have and I'm lacking time right now to look into it again.
- navigation hash updates: not sure it is worth it
- see if this can be configurable: not sure if worth it to introduce new parameter
- add links to first and last pages: done
- rebasing: I need to learn a whole new client build process and caveats, so may take some time

akash07k:
> I'm still not satisfied with what I have and I'm lacking time right now to look into it again.
> 
> * navigation hash updates: not sure it is worth it
In my opinion it will be worth it, rest @jtojnar your view on this?
> * see if this can be configurable: not sure if worth it to introduce new parameter 
> * add links to first and last pages: done
> * rebasing: I need to learn a whole new client build process and caveats, so may take some time
@jtojnar 
Can you please help?

jtojnar:
Unfortunately, I am also busy. But feel free to hit me with any questions about the client changes.

akash07k:
@niol 
@jtojnar 
Is there any progress/update on this?

jtojnar:
I will be busy until February.

akash07k:
ok bro

akash07k:
Hey @niol ,
Any updates buddy?

PR #1136, opened 2019-09-23
offline: various fixes

niol:


public/js/selfoss-db.js
```diff
@@ -705,8 +710,12 @@ selfoss.dbOffline = {
 
     sendNewStatuses: function() {
         var statuses = [];
+        var offlineStatusesCount = 0;
         selfoss.dbOffline._tr('r', selfoss.db.storage.statusq, function() {
-            selfoss.db.storage.statusq.each(function(s) {
+            selfoss.db.storage.statusq.count().then(function(c) {
```
jtojnar: Is it not possible that `selfoss.db.storage.statusq.count()` will dispatch the `then` callback after `_tr` does on line 726? I would suggest returning the count promise in this callback and taking the value in the `then` callback on 726.

```diff
--- a/public/js/selfoss-db.js
+++ b/public/js/selfoss-db.js
@@ -710,11 +710,7 @@
 
     sendNewStatuses: function() {
         var statuses = [];
-        var offlineStatusesCount = 0;
         selfoss.dbOffline._tr('r', selfoss.db.storage.statusq, function() {
-            selfoss.db.storage.statusq.count().then(function(c) {
-                offlineStatusesCount = c;
-            });
             selfoss.db.storage.statusq.limit(selfoss.dbOnline.statusUploadBatchSize).each(function(s) {
                 var statusUpdate = {
                     id: s.entryId,
@@ -723,7 +719,8 @@
                 statusUpdate[s.name] = s.value;
                 statuses.push(statusUpdate);
             });
-        }).then(function() {
+            return selfoss.db.storage.statusq.count();
+        }).then(function(offlineStatusesCount) {
             var s = undefined;
             if (statuses.length > 0) {
                 s = statuses;
```

niol: Why not but why not do that with the status array instead?

jtojnar: Yeah, those should probably do that as well. [`Promise.all`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all) will help.

niol: I find it less readable because it is required to pack both results in a an array as the unique argument to the callback...

jtojnar: Could you do it anyway? With a `TODO` comment. #1096 will allow us to use ES6 features like destructing which should make it readable again.

```javascript
selfoss.dbOffline._tr('r', selfoss.db.storage.statusq, function() {
   ...
   return Promise.all([
       count,
       statuses,
   ]);
}).then(function([count, statuses] {
    ...
});
```

jtojnar:
I have rebased this and switched the part to use async functions ([supported by `Dexie.transaction()`](https://dexie.org/docs/Dexie/Dexie.transaction()#async-and-await)) since it is even cleaner than the `Promise.all` I suggested before.

We might still want to switch to sending JSON instead of urlencoded form data (as I did in https://github.com/SSilence/selfoss/pull/1182) but this seems to work reliably enough.

Issue #1069, opened 2018-11-05
Have a way to get the current version via the API.


aminecmi:
Hi,

The master version/future release (2.19) has some breaking changes with tags handling.

As the users using my app are not using an up to date Selfoss version, having some way to simply get the current version would greatly help.

Thanks !

niol:
The version is present in the HTML head of the homepage. I can propose a pull request adding an api call, please tell me what should be implemented.

aminecmi:
@niol not sure about this, but it would be great to have some route that is accessible without login, and returns at least the version, and maybe have this route return some global settings if one day there is some.

jtojnar:
https://github.com/SSilence/selfoss/pull/1137 adds `/api/version` endpoint and starts using semver for versioning API (separately from selfoss version). Does that work for you @aminecmi?

aminecmi:
@jtojnar I'll have to test it, but that should do it. Thanks !

jtojnar:
It is now in master.

Issue #928, opened 2017-05-10
Switch to client-side templating


jtojnar:
This would make the JavaScript client equal to other clients and allow us to drop the JS-specific API endpoints entirely.

Since selfoss requires JavaScript to work and we don’t care about SEO, the only difference should be in performance. And the difference should be negligible because the majority of markup comes from the item bodies anyway.

I implemented a proof of concept in [`client-side-templates`](https://github.com/SSilence/selfoss/tree/client-side-templates) branch.

Spy-Seth:
Have you thought about making the web UI a SPA. This could make the actual selfoss API perfectly shared between all the clients (web, android, iOS). It will also make possible to make this SPA a progressive web app: this will make the mobile experience way better (if done correctly)

jtojnar:
@Spy-Seth Yes, this is the goal. niol is already working on offline support.

Issue #913, opened 2017-04-09
Cleaner file structure


jtojnar:
The existing structure is unnecessarily complicated:

1. The different source files are intermingled with the rest, all the source directories have to be set for each tool manually.
2. There are some directories that need to be accessible from the HTTP server but they are again located in different places (public and data).
3. There are directories that need to be writeable and not accessible.

### Goals
1. Related files and directories should be grouped together.
2. Files that should not be accessible should not be grouped with public ones.
3. Configuring web server should be simple – there should not be any complicated rewrite rules so that selfoss could be installed on non-traditional servers without much work.
4. Installation should be simple (`chmod`ing as few directories as possible).

### Proposed directory structure
* 📂 src
  * 📄 common.php
  * 📁 controllers
  * 📁 daos
  * 📁 helpers
  * 📁 **lang**
  * 📁 spouts
  * 📁 templates
* 📂 data (writeable)
  * 📄 **config.ini**
  * 📁 cache
  * 📁 logs
  * 📁 sqlite
* 📁 plugins
* 📂 public (accessible)
  * 📂 assets
    * 📁 css
    * 📄 favicon.svg
    * 📁 fonts
    * 📁 images
    * 📁 js
  * 📂 gen (writeable)
    * 📄 all.css
    * 📄 all.js
    * 📁 **favicons**
    * 📁 **thumbnails**
  * 📄 **index.php**
* 📄 README.md
* 📄 run.php
* 📁 vendor

The `public` directory can either be set as document root of virtual host or there can be a simple `.htaccess` in the parent directory.

### Further goals to consider
* Automated installer should be able to write `config.ini`, `data` is thus a good location for it.
* I would like to split selfoss-client off selfoss. That would mean selfoss would drop the `templates` and `assets` directories, only keeping the API, `favicons` and `thumbnails`.
* Maybe selfoss could be compiled to [PHAR](https://box-project.github.io/box2/) to reduce the number of files (faster upload to FTP). At least the `public` and `data` directories would need to be kept outside.
* If we created a plugin manager as a part of selfoss, it would need to be able to write to plugins directory.

niol:
Nice.

1. We should plan for a test directory or make the planned structure not incompatible with the presence of such directory.
2. Some people run selfoss in restricted hosting and do not have access to setting the document root: the document root is the upload root. Therefore, I think ```index.php``` in the projet root cannot be avoided, although for people selfoss on less restricted environments, putting the libs out of the document root is desirable. We should support both.

[...]

PR #609, opened 2015-01-05
allow windows 8 start screen pinning with notifications

gleroi:
This pull request add the necessary meta tags plus a method to a controller to allow windows 8 to retrieve icon and display the unread count when pinning the site on the start screen.
This should work on windows 8.1 pc/Tablette, and on windows phone 8.1.

PR #1137, opened 2019-09-25
Port to Parcel bundler

jtojnar:
Using asset bundler will allow us to switch to modern JavaScript, build user interface client-side more easily, and develop iteratively thanks to hot module replacement.

I chose Parcel over [Webpack](https://github.com/SSilence/selfoss/pull/1096) for simplicity of set up.

## What remains
- [x] ~~Determine what to do about `skipWaiting`, does the workbox-generated service worker have something similar?~~ We no longer use Workbox library.
- [ ] What is our caching strategy? Service worker should not be cached according to the comment workbox places into the file, but we do not use cache busting for it. Maybe that does not matter much, as browsers cannot keep workers more than 24 hours? We are cache busting `/api/about` requests and the HTML pages will not be cached regularly since they are returned from PHP (though service worker will still cache them, we will need to set them to network-first?).
- [x] Handle `user.{js,css}` better. Currently, we just throw `<script>` and `<link rel="stylesheet">` into the DOM, hoping that something will be loaded. I would consider creating empty files in the repository to avoid 404s but there will need to for cache to be invalidated when they are changed as well. Or can we assume user that uses such customization is familiar with `CTRL-F5`?

## What remains for future PRs
- ~~Workbox copies its files to the `public/` directory, and then imports them using `importScripts`. Would it make sense to `import` it and let Parcel bundle it?~~
- Check that babel is working correctly and that JavaScript files are properly shaken down (400 KB JS bundle is too big, https://github.com/gregtillbrook/parcel-plugin-bundle-visualiser will be helpful).
    - jQuery is full of dynamic hacks therefore hard to tree-shake and huge!
    - Stop Parcel from including huge `node-libs-browser` in `hashpassword.js` – `bcryptjs` supports using Node’s crypto library which makes Parcel think it should resolve it.
- Move API to `/api` subdirectory. This will greatly simplify the rewrite rules.
- Let web server serve the HTML files. What to do about cache invalidation? Do we need to update `.htaccess`/nginx conf…?
- Refactor to use ES classes, Promises/A+, fetch…
- Move the rest of the templates to client-side. We do not need EJS when we have JSX.
- Make more things translatable.
- Clean up the image references so that we do not have to copy those not recognized manually with `parcel-plugin-static-files-copy`. It will probably make more sense to switch directly to something like
FontAwesome, though.
- Add spinner for `/api/about` loading.
- Do not use locales in the back-end.


jtojnar:
I probably cannot make the bundling work with dynamically generated `index.html`, so I will need to add `/config` endpoint and make `home.phtml` static (which is [something I wanted to do anyway](https://github.com/SSilence/selfoss/issues/913)).

We should make sure only data relevant for public will be leaked. And include `user.js` and `user.css` there.

It might be also cleaner if we move all APIs under `/api` and while at it, we could add [`/version` endpoint](https://github.com/SSilence/selfoss/issues/1069).

jtojnar:
Do Windows still support [`msapplication-` meta tags](https://docs.microsoft.com/en-us/previous-versions/windows/internet-explorer/ie-developer/platform-apis/dn255024(v=vs.85)?redirectedfrom=MSDN). I cannot find any recent docs, only for IE. Not even for [`browserconfig.xml`](https://docs.microsoft.com/en-us/previous-versions/windows/internet-explorer/ie-developer/platform-apis/dn320426(v=vs.85)), which is still included in [h5b](https://github.com/h5bp/html5-boilerplate).

I will remove it unless someone still wants it. cc @gleroi who introduced this in https://github.com/SSilence/selfoss/pull/609

jtojnar:
With the latest push, selfoss appears to work when built with Parcel :tada: 

Here is the commit message summarizing what has been done and what is to be:

> Using asset bundler will allow us to switch to modern JavaScript, build user interface client-side more easily, and develop iteratively thanks to hot module replacement.
> 
> We chose Parcel over Webpack for simplicity of set up.
> 
> ## What changed
> 
> selfoss no longer supports run-time minification of assets. Instead, developer will need to run `npm run build` to bundle all the dependencies and our files and copy them to the `public/` directory. This means the directory is no longer needed to be writeable at run time (useful for read only locations). Alternately, developer can run `npm run dev` command, which will start a program that will watch selfoss files, rebuild them in the background when changed, and then, refresh the opened page. For common users, there are prebuilt packages available, or they can use user.{css,js} for customization, as before.
> 
> The routes `GET /` (without AJAX), `GET /password` and `GET /opml` were turned into primitive servers of static HTML files and all the magic is happening client-side. OPML is uploaded completely using XHR and even the password hash is generated client-side.
> 
> We started using some EcmaScript 6 features like imports, which Parcel will automatically resolve.
> 
> Languages are imported statically as well, to avoid having to copy them using `parcel-plugin-static-files-copy`. Unfortunately, it requires us to declare them explicitly for Parcel to be able to resolve them. And just like with the `selfoss.config` function replacing the `data` attribute lookup, I added `selfoss.lang` that reads the imported locale files directly.
> 
> We are no longer setting `<base>` in the HTML files but that was not actually necessary, as we do not use subdirectories in our front-end routes. When we start using history API, it will not be necessary either, as long as we know the routes. The base URI is therefore only necessary for cookies.
> 
> We are using `parcel-plugin-sw-cache`, which uses `workbox` to generate ServiceWorker boilerplate. Actually, we no longer write any ServiceWorker code.
> 
> 
> ## What remains
> - Determine what to do about `skipWaiting`, does the workbox-generated service worker have something similar?
> - What is our caching strategy? Service worker should not be cached according to the comment workbox places into the file, but we do not use cache busting for it. Maybe that does not matter much, as browsers cannot keep workers more than 24 hours? We are cache busting `/api/about` requests and the HTML pages will not be cached regularly since they are returned from PHP (though service worker will still cache them, we will need to set them to network-first?).
> - Handle `user.{js,css}` better. Currently, we just throw `<script>` and `<link rel="stylesheet">` into the DOM, hoping that something will be loaded. I would consider creating empty files in the repository to avoid 404s but there will need to for cache to be invalidated when they are changed as well. Or can we assume user that uses such customization is familiar with `CTRL-F5`?
> - Check that babel is working correctly and that JavaScript files are properly shaken down (400 KB JS bundle is too big, https://github.com/gregtillbrook/parcel-plugin-bundle-visualiser will be helpful).
> 
> ## What remains for future PRs
> - Workbox copies its files to the `public/` directory, and then imports them using `importScripts`. Would it make sense to `import` it and let Parcel bundle it?  https://github.com/mischnic/parcel-plugin-sw-cache/issues/22
> - Refactor to use ES classes, Promises/A+, fetch…
> - Move the rest of the templates to client-side. (#928) We do not need EJS when we have JSX.
> - Make more things translatable.
> - Clean up the image references so that we do not have to copy those not recognized manually with `parcel-plugin-static-files-copy`. It will probably make more sense to switch directly to something like FontAwesome, though. (#1013)
> - Add spinner for `/api/about` loading.
> - Do not use locales in the back-end.
> - Use [flow](https://flow.org/) annotations

gleroi:
msapplication- was introduced mainly for windows phone. With the death of the platform i think it can be safely removed.
The tag may be still supported by windows 10, but microsoft recommends to use a standard web.manifest

assets/js/selfoss-base.js
```diff
@@ -609,6 +653,19 @@ var selfoss = {
     },
 
 
+    /**
+    * Obtain a localized message for given key.
+    * @param string
+    * @return string
+    */
+    translate: function(key) {
```
niol: Why not use `selfoss.ui._()` ?

src/controllers/Index.php
```diff
@@ -91,18 +95,55 @@ public function home(Base $f3) {
...
      *
      * @return void
      */
     public function password() {
-        $this->view->password = true;
-        if (isset($_POST['password'])) {
-            $this->view->hash = hash('sha512', \F3::get('salt') . $_POST['password']);
-        }
-        echo $this->view->render('templates/hashpassword.phtml');
+        echo file_get_contents(BASEDIR . '/public/hashpassword.html');
```
niol: Maybe remove this function and serve the static file from the webserver using rewriting?

niol:
This is really interesting for a number of reasons.
I would have preferred to go with webpack which seems more popular and that would have forced me to finally learn it.
For the `skipWaiting` issue, it seems workbox build support a `skipWaiting` boolean argument that defaults to `true`. Sending `{type: 'SKIP_WAITING'}` in `selfoss-base.js` should work.

src/controllers/Index.php
jtojnar: I started with `/` and I initially wanted to do this but the fact that we use `/` for both the page and the API endpoint prevented me from doing this. Then I did not re-evaluate it for the other pages. But since `GET /` is not actually a part of public API, `GET /items` is, we probably could get away with moving it.

assets/js/selfoss-base.js
jtojnar: Yeah, that would work too. I just did not see much utility in shortening it since it will still need to be prefixed by `selfoss.ui.` for now.

The function should also accept arguments like `i18nFormat` but I wanted to only implement minimal changes to make this work.

jtojnar:
Yeah, I think we might need to go back to Webpack after all. Parcel is currently being completely rewritten to be more configurable and the plug-ins for the old version are not very flexible.

WebPack has a plug-in for Workbox straight from the [horse’s mouth](https://developers.google.com/web/tools/workbox/modules/workbox-webpack-plugin).

I really liked how Parcel had built-in support for hot-module reloading, overlay error reporting and I also felt that the HTML/CSS handling is much better in Parcel since Webpack currently only supports JavaScript and getting a CSS file is a hideous hack, but https://github.com/mischnic/parcel-plugin-sw-cache/issues/22 is a deal breaker.

jtojnar:
By the way this is the service worker which https://github.com/mischnic/parcel-plugin-sw-cache generates. `skipWaiting` is called in top-level:

```
/**
 * Welcome to your Workbox-powered service worker!
 *
 * You'll need to register this file in your web app and you should
 * disable HTTP caching for this file too.
 * See https://goo.gl/nhQhGp
 *
 * The rest of the code is auto-generated. Please don't update this file
 * directly; instead, make changes to your Workbox build configuration
 * and re-run your build process.
 * See https://goo.gl/2aRDsh
 */

importScripts("workbox-v4.3.1/workbox-sw.js");
workbox.setConfig({modulePathPrefix: "workbox-v4.3.1"});

workbox.core.skipWaiting();

workbox.core.clientsClaim();

/**
 * The workboxSW.precacheAndRoute() method efficiently caches and responds to
 * requests for URLs in the manifest.
 * See https://goo.gl/S9QRab
 */
self.__precacheManifest = [
  {
    "url": "arrow-down.fe617440.png",
    "revision": "d643bd115b8856864d69dcdd1cd64d21"
  },
  …
  {
    "url": "/",
    "revision": "0d728a7eb5884c34a6b36f32cfe8072c"
  }
].concat(self.__precacheManifest || []);
workbox.precaching.precacheAndRoute(self.__precacheManifest, {});

workbox.routing.registerNavigationRoute(workbox.precaching.getCacheKeyForURL(".//index.html"));
```

jtojnar:
The manifest inspection seems to work at least.

assets/js/selfoss-base.js
jtojnar: I completely forgot that we had `selfoss.ui._`. Especially since so many strings were obtained directly from `#lang`. Changed to that.

src/controllers/Index.php
jtojnar: One benefit of sending them through PHP is that we do not have to disable caching in the web server. To not delay this any further, I decided not to keep the current behaviour for now.

PR #1141, opened 2019-09-30
A11y: fixed aria-current for entry titles

akash07k:
Fixed aria-current for entry titles.
Previously the aria-current was getting applied to the whole entry.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -15,7 +15,7 @@
 - `logger_level=NONE` is now handled correctly ([#1077](https://github.com/SSilence/selfoss/issues/1077))
 - URLs containing special characters like commas in query string are now handled correctly ([#1082](https://github.com/SSilence/selfoss/pull/1082))
 - Set 60 second timeout to spout HTTP requests to prevent a single feed blocking other updates ([#1104](https://github.com/SSilence/selfoss/issues/1104))
-- Significantly improved accessibility ([#1133](https://github.com/SSilence/selfoss/pull/1133) and [#1134](https://github.com/SSilence/selfoss/pull/1134))
+- Significantly improved accessibility ([#1133](https://github.com/SSilence/selfoss/pull/1133), [#1134](https://github.com/SSilence/selfoss/pull/1134) and [#1141](https://github.com/SSilence/selfoss/pull/1141))
 
 ### API changes
 - `tags` attribute is now consistently array of strings, numbers are numbers and booleans are booleans. **This might break third-party clients that have not updated yet.** ([#948](https://github.com/SSilence/selfoss/pull/948))
--- a/assets/js/selfoss-shortcuts.js
+++ b/assets/js/selfoss-shortcuts.js
@@ -106,8 +106,13 @@ selfoss.shortcuts = {
                 return false;
             }
 
-            selfoss.events.entriesToolbar(selfoss.ui.entryGetSelected());
-            $('.entry.selected .entry-starr').click();
+            var selected = selfoss.ui.entryGetSelected();
+
+            if (selected !== null) {
+                selfoss.events.entriesToolbar(selected);
+                $('.entry-starr', selected).click();
+            }
+
             e.preventDefault();
             return false;
         });
@@ -118,8 +123,13 @@ selfoss.shortcuts = {
                 return false;
             }
 
-            selfoss.events.entriesToolbar(selfoss.ui.entryGetSelected());
-            $('.entry.selected .entry-unread').click();
+            var selected = selfoss.ui.entryGetSelected();
+
+            if (selected !== null) {
+                selfoss.events.entriesToolbar(selected);
+                $('.entry-unread', selected).click();
+            }
+
             e.preventDefault();
             return false;
         });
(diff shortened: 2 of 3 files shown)


public/js/selfoss-ui.js
```diff
@@ -91,6 +91,7 @@ selfoss.ui = {
 
         entry.addClass('expanded');
         $('.entry-title > .entry-title-link', entry).attr('aria-expanded', 'true');
+        $('.entry-title > .entry-title-link', entry).attr('aria-current', 'true');
```
jtojnar: This will mark every opened entry as the current one.

public/js/selfoss-ui.js
```diff
@@ -176,13 +178,11 @@ selfoss.ui = {
     entrySelect: function(entry) {
         if (selfoss.ui.selectedEntry !== null) {
             selfoss.ui.selectedEntry.removeClass('selected');
-            selfoss.ui.selectedEntry.attr('aria-current', 'false');
```
jtojnar: Instead you need to mark the title of selected entry here: `$('.entry-title > .entry-title-link', selfoss.ui.selectedEntry).attr(...)`

akash07k: > Instead you need to mark the title of selected entry here: `$('.entry-title > .entry-title-link', selfoss.ui.selectedEntry).attr(...)`

Done the changes. Please check. However, I'm yet to test whether it works properly or not.

public/js/selfoss-ui.js
```diff
@@ -176,13 +176,13 @@ selfoss.ui = {
...
+        $('.entry-title > .entry-title-link', entry).attr('aria-current', 'false');
```
jtojnar: At least this one should refer to the old `selectedEntry`. `entry` is the one we want to select. 
```suggestion
            $('.entry-title > .entry-title-link', selfoss.ui.selectedEntry).attr('aria-current', 'false');
```

public/js/selfoss-ui.js
```diff
@@ -176,13 +176,13 @@ selfoss.ui = {
...
         }
 
         selfoss.ui.selectedEntry = entry;
 
         if (entry) {
-            selfoss.ui.selectedEntry.attr('aria-current', 'true');
+        $('.entry-title > .entry-title-link', entry).attr('aria-current', 'true');
```
jtojnar: Indentation fix
```suggestion
            $('.entry-title > .entry-title-link', entry).attr('aria-current', 'true');
```

public/js/selfoss-ui.js
akash07k: > At least this one should refer to the old `selectedEntry`. `entry` is the one we want to select.

I've done the changes, but I don't think so that it is working appropriately. Now, even though I'm collapsing the entry, Screen reader is still announcing 'current' for the entry. it mean's that aria-current="false" is not invoking. Having said that, aria-current is only going away from the previous entry if I'm expanding another entry. and again, same is happening with it as well after collapsing. aria-current is still true.

akash07k: > At least this one should refer to the old `selectedEntry`. `entry` is the one we want to select.

I feel that 'selfoss.ui.selectedEntry' is not needed for aria-current

jtojnar: > Now, even though I'm collapsing the entry, Screen reader is still announcing 'current' for the entry.

That is correct behaviour. The last entry you interacted with will be selected (visually, there is a darker border around it). Entry being selected means you can use selfoss keyboard shorcuts to expand it again (<kbd>o</kbd>), open the link (<kbd>v</kbd>), or switch to a previous or next entry (<kbd>j</kbd> and <kbd>k</kbd>, or arrow keys).

akash07k:
Oh, ok, it's great then, but those keys are not working and even
screen reader is not announcing them.
seems that I'll have to look at it as well then. :-(

jtojnar:
Hmm, I thought I tested that edge case but apparently not. Will push a fix shortly.

jtojnar:
Is this good to go?

jtojnar:
Rebased onto master for clean git history.

akash07k:
> Is this good to go?

Yes, it is fine now, but still the keys are not being identified by the screen readers, I feel that I'll have to look deep down into it and will have to manually expose the commands by some methods. But as of now, we can leave it aside for future.

jtojnar:
Yeah, they are not very visible for sighted users either. User will either already know the keyboard shortcuts from Google Reader, which we inherited, or read about them on the web page. We should add a [help overlay](https://github.com/SSilence/selfoss/issues/1143) when pressing <kbd>h</kbd>.

Not sure how to expose the shortcuts to visually impaired users other than adding buttons for every action there is a shortcut for, and putting `aria-keyshortcuts` to that. But I am not sure how usable it would be with the 22 shortcuts selfoss currently supports.

akash07k:
I think that something is better than nothing.
Although the user will have to remember all the shortcuts, but
eventually they will be able to become habitual to these shortcuts.
using aria-keyshortcuts is a good idea indeed.
But implementing overlay will be tricky I feel

PR #1142, opened 2019-10-02
Share button for copy

### LINKED ISSUE #1105: Review of current share button
Hi, here is a quick review of the "share" buttons:
image

We have:

- Google+ => but Google+ is dead since since april 2019 (https://plus.google.com/)
- Twitter, Facebook, Pocket, Delicious, Mail =>

I think that we can remove Google+ and instead add a button "copy to clipboard". This new button will copy the link to the original article in clipboard. Then users can paste the link every where : in a document, in a mail, in a messenger ans so on. What do you think ?

MiteyIronPaw:
Fixes #1105 

- Still needs Translations.
- It might be worth making the pop-up at the top have absolute positioning so that it can be seen when the user scrolls down and a high z-index so that mobile users can see it when they have a item open.

Otherwise should be good to go. Tested on Firefox, Chrome both on Linux desktop and android.

### DIFF
--- a/_docs/website/index.html
+++ b/_docs/website/index.html
@@ -288,7 +288,19 @@ db_port=3306</code>
                     </tr>
                     <tr>
                         <td class="documentation-first-column">share</td>
-                        <td>share defines which share buttons beneath the entry are visible. Default is share=tfpde (f = facebook, t = twitter, p = pocket, d = diaspora, w = wallabag, e = email). When you would like to show only facebook and twitter share button use share=ft.</td>
+                        <td>
+                            <p><code>share</code> defines which sharing buttons beneath the entry are visible. The following methods are supported:</p>
+                            <dl>
+                                <dt>f</dt><dd>Facebook</dd>
+                                <dt>t</dt><dd>Twitter</dd>
+                                <dt>p</dt><dd>Pocket</dd>
+                                <dt>d</dt><dd>Diaspora</dd>
+                                <dt>w</dt><dd>Wallabag</dd>
+                                <dt>e</dt><dd>E-Mail</dd>
+                            </dl>
+                            <p>Include the letters for methods you want to use. For example, if you would like to only show Facebook and Twitter share buttons, use <code>share=ft</code>.</p>
+                            <p>Defaults to <code>share=tfpde</code>.</p>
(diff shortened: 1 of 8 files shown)


public/js/selfoss-base.js
```diff
@@ -621,6 +621,47 @@ var selfoss = {
...
+
+
+    /**
+     * Copy a string to the users clipboard.
+     * This will deselect anything the user has selected and overwirte their clipboard.
+     *
+     * @param stringToCopy
+     * @return true if no execptions were thrown
+     */
+    copyString: function(stringToCopy) {
```
jtojnar: Note, once we merge #1137, we will be able to use something like https://www.npmjs.com/package/copy-to-clipboard.

public/js/selfoss-shares.js
```diff
@@ -33,6 +33,10 @@ selfoss.shares = {
         this.register('mail', 'e', function(url, title) {
             document.location.href = 'mailto:?body=' + encodeURIComponent(url) + '&subject=' + encodeURIComponent(title);
         });
+        this.register('copy', 'c', function(url) {
+            selfoss.copyString(url);
+            selfoss.ui.showMessage(selfoss.ui._('url_copied'));
```
jtojnar: We currently do not have `selfoss.ui._`. #1137 introduces `selfoss.ui.translate`, though.

jtojnar: I retract this comment. Forgot we had this.

public/js/selfoss-base.js
MiteyIronPaw: Thoughts on merging #1137 into this branch then blocking it until #1137 has been merged?

jtojnar: 👍Hopefully, it will not take much longer to merge it.

jtojnar: Yay, it is in master now.

jtojnar:
Thanks for the help. I switched to a different clipboard library that is closer to the native version. Also changed the icon to a vector one since #1013 got finally merged.

MiteyIronPaw:
Haha thank you :) , I didn't end up doing much in the end.

Issue #1144, opened 2019-10-03
OPML Page is throwing errors


akash07k:
I'm using the latest master branch and after todays updates of webpack, OPML page is not working.
only one error is there in logs.
Below is my error log:
[2019-10-03 20:16:41] selfoss.ERROR: readfile(/home/akash/public_html/readerotica/src/../public/opml.html): failed to open stream: No such file or directory  
[2019-10-03 20:17:08] selfoss.ERROR: readfile(/home/akash/public_html/readerotica/src/../public/opml.html): failed to open stream: No such file or directory  
[2019-10-03 20:17:11] selfoss.ERROR: readfile(/home/akash/public_html/readerotica/src/../public/opml.html): failed to open stream: No such file or directory  
[2019-10-03 23:11:09] selfoss.ERROR: readfile(/home/akash/public_html/readerotica/src/../public/opml.html): failed to open stream: No such file or directory

jtojnar:
Did you run `npm run build`?

akash07k:
> Did you run `npm run build`?

Oh, I was not aware about it. ran it and now everything is fine. closing it. So, Will I have to build everytime when there's a change in js files?

jtojnar:
If you obtain a prebuilt zip archive (one of the first two options on https://github.com/SSilence/selfoss#download), you should be good to go. similarily, if you use selfoss package from a Linux distribution, building the assets should be taken care of. If you use selfoss from git, though, you will indeed need to build it whenever the assets change.

But apparently, the CI does not include the public directory into the zip. Which I did not notice since I had the files already built. Will fix that shortly.

akash07k:
> If you obtain a prebuilt zip archive (one of the first two options on https://github.com/SSilence/selfoss#download), you should be good to go. similarily, if you use selfoss package from a Linux distribution, building the assets should be taken care of. If you use selfoss from git, though, you will indeed need to build it whenever the assets change.
> 
> But apparently, the CI does not include the public directory into the zip. Which I did not notice since I had the files already built. Will fix that shortly.

so what should be the correct sequence of the commands? first npm run install-dependencies
then,
npm npm run --prefix assets/ build

jtojnar:
Weird, I have thought I included it in the README, but I probably forgot about it. Yes, `npm run install-dependencies` and then `npm run build` (no need to specify `--prefix,` as we redirect the `dev` and `build` commands from the root to assets automatically).

jtojnar:
[...]

PR #1145, opened 2019-10-03
Fix dist

jtojnar:
Two regressions from #1137.

Closes: https://github.com/SSilence/selfoss/issues/1144


akash07k:
So is it merged in master?

akash07k:
O, just checked, it is not merged. I'll apply the patch

jtojnar:
It was merged to master, not sure what you saw.

akash07k:
Ya, realised the same.
Actually there was some problem with my caches. when I cleared and
refreshed the pages, I found it got merged

Issue #1161, opened 2019-12-22
How to dockerize Selfoss


akash07k:
I want to dockerize selfoss, but I'm unable to identify that how exactly I do it since it uses PHP and nodejs both.
If I require to use docker-compose, then can someone please help me with the exact proper script?

jtojnar:
That will depend on what you want to use the Docker image for.

If you want to use if for running selfoss on production, you do not need node, you can use one of the prebuilt archives in https://github.com/SSilence/selfoss#download and copy one of the existing Dockerfiles from Docker Hub:

* https://hub.docker.com/r/hardware/selfoss/dockerfile
* https://hub.docker.com/r/pamplemousse/selfoss/dockerfile

If you want to use it for development, I do not think docker-compose is necessary for node. The JavaScript assets are served by the web server so I would expect them to reside in the same image. You could install Node in the Dockerfile just like PHP and other stuff is.

If you want to use database other than SQLite, then docker-compose might be useful to orchestrate the communication between the PHP server and the DB server but it is not necessary either. Or if you want to use PHP-FPM.

squatica:
I'm actually currently preparing a Dockerfile for selfoss with the intention to send a PR. As @jtojnar said there are two main scenarios - you either want the "production" image which contains app sources inside, or a "dev" image which only contains dev dependencies and tools, but sources are mounted to your filesystem.
This is what I have so far for the production image (uses the stable tag 2.18), dev image coming soon.
```
FROM ubuntu:bionic as source

RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install --no-install-recommends -y \
    git ca-certificates \
    && git clone https://github.com/SSilence/selfoss.git && cd selfoss \
    && git checkout 2.18



FROM composer:1.9 as composer

COPY --from=source /selfoss /selfoss

RUN cd /selfoss && composer install --ignore-platform-reqs --optimize-autoloader --no-dev



FROM node:12-stretch as npm

COPY --from=composer /selfoss /selfoss

RUN cd /selfoss && npm i && ./node_modules/.bin/grunt client:install



FROM php:7.3-apache-stretch

COPY --from=npm /selfoss /var/www/html/

RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install --no-install-recommends -y libpng-dev \
    && docker-php-ext-install -j$(nproc) gd \
    && apt-get clean && rm -rf /var/lib/apt/lists/*

WORKDIR /var/www/html/

[...]

PR #1162, opened 2019-12-26
Added the docker support

akash07k:
Added the docker support
original credit goes to @squatica. <https://github.com/squatica/selfoss-docker>
I updated it to work with the selfoss repo regardless of the local changes and also tweaked some things.

Dockerfile for Selfoss RSS aggregator is bundled with the repository.

Selfoss config is mounted in a separate volume, so your custom settings should survive reboot.

To run the latest stable tag (2.18) use:
```
docker-compose up
```
Then find the web interface at http://localhost:8390

To run the latest master version use:
```
docker-compose -f docker-compose-master.yml up
```
Then find the web interface at http://localhost:8391

To rebuild the master version with the latest code from git use:
```
docker-compose -f docker-compose.master.yml build --no-cache
```


jtojnar:
I am not all that familiar with Docker, how do other web apps handle it? Especially the distinction between dev image and production image – there should be an explanation of what the Docker images are intended for and how should they be used.

Also I do not think it makes sense to duplicate `Dockerfile` for latest stable version in the repo. User could just `git checkout <latest stable version>` and use `Dockerfile` from there.

It would be cleaner to move the Docker-related files to somewhere like `utils/docker`, the root is already crowded as is.

Also I read somewhere in the past that multiple `FROM`s are bad. Is that not a case any more?

akash07k:
> I am not all that familiar with Docker, how do other web apps handle it? Especially the distinction between dev image and production image – there should be an explanation of what the Docker images are intended for and how should they be used.
-- I'll check regarding the explainations in the comments of the dockerfile.
> Also I do not think it makes sense to duplicate `Dockerfile` for latest stable version in the repo. User could just `git checkout <latest stable version>` and use `Dockerfile` from there.
-- I'm also thinking to use only 1 dockerfile. will remove the Dockerfile responsible for checking out the stable version.
> It would be cleaner to move the Docker-related files to somewhere like `utils/docker`, the root is already crowded as is.
 I agree with the root croud issue, but moving the dockerfiles to the utils folder can be confusing for some people and will increase the complications. I don't think that we should do it. there will only be 3 extra files with the latest changes after the removal of the 1 dockerfile and corresponding docker-compose file. Those 3 will be: 'Dockerfile', 'docker-compose.yml' and '.dockerignore. I feel that having only 3 extra files for the sake of native docker compattibility is not a big deal.
> Also I read somewhere in the past that multiple `FROM`s are bad. Is that not a case any more?

-- No, now docker has inbuilt support for multiple base images I.E: multiple froms.

akash07k:
Fixes
#1161

akash07k:
> I am not all that familiar with Docker, how do other web apps handle it? Especially the distinction between dev image and production image – there should be an explanation of what the Docker images are intended for and how should they be used.
> 
> Also I do not think it makes sense to duplicate `Dockerfile` for latest stable version in the repo. User could just `git checkout <latest stable version>` and use `Dockerfile` from there.
> 
> It would be cleaner to move the Docker-related files to somewhere like `utils/docker`, the root is already crowded as is.
> 
> Also I read somewhere in the past that multiple `FROM`s are bad. Is that not a case any more?
Please refer:
https://docs.docker.com/develop/develop-images/multistage-build/
https://dev.to/brpaz/using-docker-multi-stage-builds-during-development-35bc

.dockerignore
```diff
@@ -0,0 +1,16 @@
...
+/data/logs/*.log
+/data/sqlite/*.db
+/public
+user.css
+user.js
+config.ini
+node_modules
+.env
+vendor/
+.php_cs.cache
```
jtojnar: According to https://codefresh.io/docker-tutorial/not-ignore-dockerignore-2/, we should also add `.git`.

Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+
+RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install --no-install-recommends -y libpng-dev \
+    && docker-php-ext-install -j$(nproc) gd \
+    && apt-get clean && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /var/www/html/
+
+RUN mkdir config && ln -s config/config.ini config.ini \
+    && a2enmod rewrite \
+    && /bin/echo -e "#!/bin/bash\nchown -R www-data:www-data /var/www/html/data/cache /var/www/html/data/favicons /var/www/html/data/logs /var/www/html/data/thumbnails /var/www/html/data/sqlite" > /entrypoint.sh \
```
jtojnar: I find this ugly. Could we create `utils/docker/entrypoint.sh` and call that here?

README.md
```diff
@@ -77,6 +77,25 @@ If you want to create a package with all the dependencies bundled, you can run `
 
 Every patch is expected to adhere to our coding style, which is checked automatically by Travis. You can install the checkers locally using `npm run install-dependencies`, and then run the checks using `npm run check` before submitting a pull request. There is also `npm run fix`, that will attempt to fix the formatting.
 
+## Dockerizing selfoss
+
+Dockerfile for Selfoss RSS aggregator is bundled in the repository.
+
+Selfoss config is mounted in a separate volume, so your custom settings should survive reboot.
```
jtojnar: Is that property of Docker or docker-compose?

.dockerignore
jtojnar: Also probably `/utils`.

akash07k: > According to https://codefresh.io/docker-tutorial/not-ignore-dockerignore-2/, we should also add `.git`.

Updated.

akash07k: > Also probably `/utils`.

You mean we should exclude /utils too from .dockerignore?

Dockerfile
akash07k: > I find this ugly. Could we create `utils/docker/entrypoint.sh` and call that here?

Yes, it's a good idea but unfortunately, I'm not that good with bash/shell scripting :(

README.md
akash07k: > Is that property of Docker or docker-compose?

Yes, we can mount volumes via docker

.dockerignore
jtojnar: Yeah, I actually have ton of uncommitted stuff in `/utils`. Though we would also need `!/utils/docker` rule.

README.md
jtojnar: But does not that need some extra CLI flags passed to `docker`?

Dockerfile
```diff
@@ -0,0 +1,44 @@
+FROM ubuntu:bionic as source
+
+RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install --no-install-recommends -y \
```
jtojnar: Is this stage still necessary?

Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+
+RUN cd /selfoss && composer install --ignore-platform-reqs --optimize-autoloader --no-dev
+
+
+
+FROM node:13-buster as npm
+
+COPY --from=composer /selfoss /selfoss
+
+RUN cd /selfoss && npm i && npm install --prefix assets/ && npm run build
```
jtojnar:
```suggestion
RUN cd /selfoss && npm install && npm install --prefix assets/ && npm run build
```

Though the top-level `npm install` is only used for creating a dist tarball so maybe we can omit it altogether.

```suggestion
RUN cd /selfoss && npm install --prefix assets/ && npm run build
```

Dockerfile
jtojnar: I think just literally calling the command should work:

```
mkdir utils/docker \
&& echo -e "#!/bin/bash\nchown -R www-data:www-data /var/www/html/data/cache /var/www/html/data/favicons /var/www/html/data/logs /var/www/html/data/thumbnails /var/www/html/data/sqlite" > utils/docker/entrypoint.sh
&& echo -e "if [ ! -f config/config.ini ]; then cp defaults.ini config/config.ini && sed -i 's/logger_destination=.*$/logger_destination=file:php:\/\/stderr/' config/config.ini; fi" >> utils/docker/entrypoint.sh \
&& echo -e "su www-data -s /bin/bash -c 'php /var/www/html/cliupdate.php' >/dev/null 2>&1" >> utils/docker/entrypoint.sh \
&& echo -e "(while true; do su www-data -s /bin/bash -c 'php /var/www/html/cliupdate.php'; sleep 900; done;) &" >> utils/docker/entrypoint.sh \
&& echo -e "apache2-foreground" >> utils/docker/entrypoint.sh \
&& chmod a+x utils/docker/entrypoint.sh
```

Plus maybe some clean up after that.

Dockerfile
akash07k: > Though the top-level `npm install` is only used for creating a dist tarball so maybe we can omit it altogether.

Done the updates

akash07k: > Though the top-level `npm install` is only used for creating a dist tarball so maybe we can omit it altogether.

So, we can remove the npm install part altogether? please correct me if I'm wrong because I didn't get it properly

Dockerfile
akash07k: > Is this stage still necessary?

No, removed it.
Added the apt-get upgrade so that the packages can be updated before commiting the image.

README.md
```diff
@@ -77,6 +77,25 @@ If you want to create a package with all the dependencies bundled, you can run `
 
 Every patch is expected to adhere to our coding style, which is checked automatically by Travis. You can install the checkers locally using `npm run install-dependencies`, and then run the checks using `npm run check` before submitting a pull request. There is also `npm run fix`, that will attempt to fix the formatting.
 
+## Dockerizing selfoss
+
+Dockerfile for Selfoss RSS aggregator is bundled in the repository.
```
jtojnar: It is not clear to me from this text who this is directed for. Is this for:

1. People who want to run production instances of selfoss but want it isolated from other apps on their system. — Docker might not be the best way to do that but perhaps it is the most convenient one.
2. People who want to run production selfoss and do not want to bother with installing web server and PHP (they do not need to bother with composer and NPM as we ship prebuilt archives) – this looks like the most probable option. 
3. Developers who do not want to bother with installing web server, composer and NPM — fair but developers will probably also need to run `npm run build` or `composer install` when they change something, is there a way to do that without rebuilding the whole image? Also `npm run dev`.
4. Developers that want uniform development environment – that does not make much sense since selfoss is expected to run at wide variety of web servers and PHP runtimes.

jtojnar: The `README` should explain which of these use cases are targeted and how to accomplish their goals.

Dockerfile
akash07k: > I think just literally calling the command should work:
> 
> ```
> mkdir utils/docker \
> && echo -e "#!/bin/bash\nchown -R www-data:www-data /var/www/html/data/cache /var/www/html/data/favicons /var/www/html/data/logs /var/www/html/data/thumbnails /var/www/html/data/sqlite" > utils/docker/entrypoint.sh
> && echo -e "if [ ! -f config/config.ini ]; then cp defaults.ini config/config.ini && sed -i 's/logger_destination=.*$/logger_destination=file:php:\/\/stderr/' config/config.ini; fi" >> utils/docker/entrypoint.sh \
> && echo -e "su www-data -s /bin/bash -c 'php /var/www/html/cliupdate.php' >/dev/null 2>&1" >> utils/docker/entrypoint.sh \
> && echo -e "(while true; do su www-data -s /bin/bash -c 'php /var/www/html/cliupdate.php'; sleep 900; done;) &" >> utils/docker/entrypoint.sh \
> && echo -e "apache2-foreground" >> utils/docker/entrypoint.sh \
> && chmod a+x utils/docker/entrypoint.sh
> ```
> 
> Plus maybe some clean up after that.

Where should I call this command? in Dockerfile or on my local system to create the entrypoint.sh?

README.md
akash07k: > But does not that need some extra CLI flags passed to `docker`?

If we use docker-compose then I think no

README.md
akash07k: > It is not clear to me from this text who this is directed for. Is this for:
> 
> 1. People who want to run production instances of selfoss but want it isolated from other apps on their system. — Docker might not be the best way to do that but perhaps it is the most convenient one.
> 2. People who want to run production selfoss and do not want to bother with installing web server and PHP (they do not need to bother with composer and NPM as we ship prebuilt archives) – this looks like the most probable option.
> 3. Developers who do not want to bother with installing web server, composer and NPM — fair but developers will probably also need to run `npm run build` or `composer install` when they change something, is there a way to do that without rebuilding the whole image? Also `npm run dev`.
> 4. Developers that want uniform development environment – that does not make much sense since selfoss is expected to run at wide variety of web servers and PHP runtimes.

Running selfoss on docker is on user's choice.
If there's some person who wants to run selfoss on docker then probably he/she will know their use case.
 I think defining/writing it will be too fffortful and time consuming as well.

Dockerfile
jtojnar: Does anything of the stage get carried into the next one? You are only copying `.` and no of the commands should write there.

README.md
akash07k: > It is not clear to me from this text who this is directed for. Is this for:
> 
> 1. People who want to run production instances of selfoss but want it isolated from other apps on their system. — Docker might not be the best way to do that but perhaps it is the most convenient one.
> 2. People who want to run production selfoss and do not want to bother with installing web server and PHP (they do not need to bother with composer and NPM as we ship prebuilt archives) – this looks like the most probable option.
> 3. Developers who do not want to bother with installing web server, composer and NPM — fair but developers will probably also need to run `npm run build` or `composer install` when they change something, is there a way to do that without rebuilding the whole image? Also `npm run dev`.
--It may be tricky and we'll have to work on this usecase that is, if the source gets updated then use npm run build or npm run dev.
> 4. Developers that want uniform development environment – that does not make much ...

Dockerfile
jtojnar: Well, the top-level `npm install` only installs `archiver` needed for `npm run dist`:

https://github.com/SSilence/selfoss/blob/cdde3b5affb931d360f4952b3f2b6d15cdaf467a/package.json#L9-L11

But `npm` also runs `postinstall` script:

https://github.com/SSilence/selfoss/blob/cdde3b5affb931d360f4952b3f2b6d15cdaf467a/package.json#L32

which will install the dependencies:

https://github.com/SSilence/selfoss/blob/cdde3b5affb931d360f4952b3f2b6d15cdaf467a/package.json#L27-L29

So the first `install` will install `archiver` that we do not need but also run the second `install` so we are running it twice. And it will also run `composer install`, I wonder how it does not fail.

Dockerfile
jtojnar: On the local system. It should create the file on your local system and you can then just call `utils/docker/entrypoint.sh` in the `Dockerfile`.

README.md
jtojnar: Yeah, then I think it should be made clear that it is docker-compose that creates the volumes. Or show how to do it using plain `docker` invocation.

jtojnar:
On a side note, could you please insert blank line between quotations and your replies? Unfortunately, GitHub will incorporate the reply into the blockquote if you don’t, making it very hard to read.

akash07k:
> On a side note, could you please insert blank line between quotations and your replies? Unfortunately, GitHub will incorporate the reply into the blockquote if you don’t, making it very hard to read.

Sure, I'll take care of that now.

Dockerfile
akash07k: > Well, the top-level `npm install` only installs `archiver` needed for `npm run dist`:
> 
> https://github.com/SSilence/selfoss/blob/cdde3b5affb931d360f4952b3f2b6d15cdaf467a/package.json#L9-L11
> 
> But `npm` also runs `postinstall` script:
> 
> https://github.com/SSilence/selfoss/blob/cdde3b5affb931d360f4952b3f2b6d15cdaf467a/package.json#L32
> 
> which will install the dependencies:
> 
> https://github.com/SSilence/selfoss/blob/cdde3b5affb931d360f4952b3f2b6d15cdaf467a/package.json#L27-L29
> 
> So the first `install` will install `archiver` that we do not need but also run the second `install` so we are running it twice. And it will also run `composer install`, I wonder how it does not fail.

Not sure. we are not running run npm install-dependencies because it is failing.
Not only on docker, even on my LAMP server which is based on Ubuntu 18.04 and on XAMPP also.
I've checked it from master branch too, but there also the same issue.
in october it aas working fine.

Dockerfile
akash07k: > Does anything of the stage get carried into the next one? You are only copying `.` and no of the commands should write there.

Hmm, right.
As per my understanding, nothing is being carried over.
but I'm not sure.
@squaticaj can you please help here?

README.md
akash07k: > Yeah, then I think it should be made clear that it is docker-compose that creates the volumes. Or show how to do it using plain `docker` invocation.

Sure, I'll do it, and also, I'll add the volume support in the dockerfile itself. then it won't require anything additional via docker run also.

Dockerfile
akash07k: > On the local system. It should create the file on your local system and you can then just call `utils/docker/entrypoint.sh` in the `Dockerfile`.

Ok, going to try that.

akash07k: > On the local system. It should create the file on your local system and you can then just call `utils/docker/entrypoint.sh` in the `Dockerfile`.

Ok, yes, the command works and the file is created.
can you please check it?
If possible then can you please add the required cleanups which you were talking about?
(I've changed the extention for now because github wasn't allowing the .sh uploads)

jtojnar: This looks more usable:

```bash
#!/bin/bash

# Make data directories owned by Apache
chown -R www-data:www-data \
    /var/www/html/data/cache \
    /var/www/html/data/favicons \
    /var/www/html/data/logs \
    /var/www/html/data/thumbnails \
    /var/www/html/data/sqlite

# Create a config file when one does not exist
if [[ ! -f config/config.ini ]]; then
    cp defaults.ini config/config.ini \
    && sed -i 's#^logger_destination=.*#logger_destination=file:php://stderr#' config/config.ini
fi

# Run updater process periodically
su www-data -s /bin/bash -c 'php /var/www/html/cliupdate.php' >/dev/null 2>&1
(while true; do su www-data -s /bin/bash -c 'php /var/www/html/cliupdate.php'; sleep 900; done;) &

# Start the server
apache2-foreground
```

Though I am not sure about several things:
* Does the owner need to be changed? How does that translate across the mounted volume? What if www-data (or user with the same uid) does not exist outside of Docker container?
* Do we need to create the config file when it does not exists? selfoss will happily read the data from `defaults.ini` when some keys or even the whole `config.ini` is missing.
    * Maybe discoverability is a boon here, since `defaults.ini` will be hidden inside the container.
* Do we need to set the logger destination in the config, or might it make better sense forcing it through `SELFOSS_LOGGER_DESTINATION` environment variable.
* The while loop inside subprocess is very ugly. Does docker have some mechanism for running commands periodically like cron or systemd timers? I found https://developers.redhat.com/blog/2019/04/24/how-to-run-systemd-in-a-container/ but do not have time to study it in detail.

Dockerfile
jtojnar: > Not sure. we are not running run npm install-dependencies because it is failing.
> Not only on docker, even on my LAMP server which is based on Ubuntu 18.04 and on XAMPP also.
> I've checked it from master branch too, but there also the same issue.
> in october it aas working fine.

Could you share the error message? It works for me here on master.

akash07k: > > Not sure. we are not running run npm install-dependencies because it is failing.
> > Not only on docker, even on my LAMP server which is based on Ubuntu 18.04 and on XAMPP also.
> > I've checked it from master branch too, but there also the same issue.
> > in october it aas working fine.
> 
> Could you share the error message? It works for me here on master.

Sure, here it is:
(I don't know whether the screenshot is showing properly or not as I'm a visually impaired user). If it doesn't display properly, then please let me know
image

Dockerfile
akash07k: > This looks more usable:
> 
> ```shell
> #!/bin/bash
> 
> # Make data directories owned by Apache
> chown -R www-data:www-data \
>     /var/www/html/data/cache \
>     /var/www/html/data/favicons \
>     /var/www/html/data/logs \
>     /var/www/html/data/thumbnails \
>     /var/www/html/data/sqlite
> 
> # Create a config file when one does not exist
> if [[ ! -f config/config.ini ]]; then
>     cp defaults.ini config/config.ini \
>     && sed -i 's#^logger_destination=.*#logger_destination=file:php://stderr#' config/config.ini
> fi
> 
> # Run updater process periodically
> su www-data -s /bin/bash -c 'php /var/www/html/cliupdate.php' >/dev/null 2>&1
> (while true; do su www-data -s /bin/bash -c 'php /var/www/html/cliupdate.php'; sleep 900; done;) &
> 
> # Start the server
> apache2-foreground
> ```

> Ok, I'll create the entrypoint.sh from this text.


> 
> Though I am not sure about several things:
> 

> > * Does the owner need to be changed? How does that translate across the mounted volume? What if www-data (or user with the same uid) does not exist outside of Docker container?

> Not sure about it, @squatica can you please help here?



> * Do we need to create the config file when it does not exists? selfoss will happily read the data from `defaults.ini` when some keys or even the whole `config.ini` is missing.
>   
> I think no, but I don't know why, it is still not taking any configurations from config.ini while running on docker, I think I'll have to investigate more on this.

>   * Maybe discoverability is a boon here, since `defaults.ini` will be hidden inside the container.
> * Do we need to set the logger destination in the config, or might it make better sense forcing it through `SELFOSS_LOGGER_DESTINATION` environment variable.

> We can any of the option from above, what do you recommend? should we use environment variables?

> * The while loop inside subprocess is very ugly. Does docker have some mechanism for running commands periodically like cron or systemd timers? I found https://developers.redhat.com/blog/2019/04/24/how-to-run-systemd-in-a-container/ but do not have time to study it in detail.


> Actually we'll have to install crontab inside the container base image, I'll take a look at it soon.

> This looks more usable:
> 
> ```shell
> #!/bin/bash
> 
> # Make data directories owned by Apache
> chown -R www-data:www-data \
>     /var/www/html/data/cache \
>     /var/www/html/data/favicons \
>     /var/www/html/data/logs \
>     /var/www/html/data/thumbnails \
>     /var/www/html/data/sqlite
> 
> # Create a config file when one does not exist
> if [[ ! -f config/config.ini ]]; then
>     cp defaults.ini config/config.ini \
>     && sed -i 's#^logger_destination=.*#logger_destination=file:php://stderr#' config/config.ini
> fi
> 
> # Run updater process periodically
> su www-data -s /bin/bash -c 'php /var/www/html/cliupdate.php' >/dev/null 2>&1
> (while true; do su www-data -s /bin/bash -c 'php /var/www/html/cliupdate.php'; sleep 900; done;) &
> 
> # Start the server
> apache2-foreground
> ```
> 
> Though I am not sure about several things:
> 
> * Does the owner need to be changed? How does that translate across the mounted volume? What if www-data (or user with the same uid) does not exist outside of Docker container?
> * Do we need to create the config file when it does not exists? selfoss will happily read the data from `defaults.ini` when some keys or even the whole `config.ini` is missing.
>   
>   * Maybe discoverability is a boon here, since `defaults.ini` will be hidden inside the container.
> * Do we need to set the logger destination in the config, or might it make better sense forcing it through `SELFOSS_LOGGER_DESTINATION` environment variable.
> * The while loop inside subprocess is very ugly. Does docker have some mechanism for running commands periodically like cron or systemd timers? I found https://developers.redhat.com/blog/2019/04/24/how-to-run-systemd-in-a-container/ but do not have time to study it in detail.

Dockerfile
jtojnar: This was from calling `npm run install-dependencies`? The only thing I see is the fist line: Missing package version could you try adding `version` attribute to assets/package.json. Though I wonder why does not trip it my npm of the almost same version (6.13.0).

akash07k: > This was from calling `npm run install-dependencies`? The only thing I see is the fist line: Missing package version could you try adding `version` attribute to assets/package.json. Though I wonder why does not trip it my npm of the almost same version (6.13.0).

Yes, I ran npm run install-dependencies. so, should I add npm package version to assets/package.json?

.dockerignore
squatica: We need to make a clear distinction about the purpose of this docker image. Two main types are:
- production
- development

Where the production can further be devided into:
- sources come from git clone
- sources come from the current directory (`ADD . /src` in Dockerfile)

My Dockerfile was intended as the production image with sources from git clone. Here is why: it can then be in a separate git repo (such as github.com/SSilence/selfoss-docker) and can be built completely separately yet always resulting in the same source files. It can also be built for some older tag, which is good either for comparison or when master is broken. Using production image with `ADD . /src` is problematic because then even your uncommitted files will be included in the image, which is a problem because you may accidentally push into docker hub something that doesn't exist anywhere in git - this is also how viruses get into docker images. In practice this approach is only used when the build is automated and done in a pipeline on the build server, where we can make sure it checks out a clean workspace from git before the docker build.

So if we keep this image as the production image with sources from git, we don't even need any .dockerignore file at all, because the build will run in a different directory - different git repo.

Dockerfile
```diff
@@ -0,0 +1,46 @@
+FROM ubuntu:bionic as source
+
+# Set frontend mode as noninteractive (default answers to all questions)
+ENV DEBIAN_FRONTEND noninteractive
```
squatica: I like to use DEBIAN_FRONTEND=noninteractive directly on the line before apt-get install, because if it's configured like this globally at the beginning, it will remain set even when the container is running, so even if you later `docker exec` into the container and want to `apt-get install` something, apt will think you are still non-interactive.

Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+
+RUN apt-get update &&\
+    apt-get upgrade -y && \
```
squatica: apt-get upgrade in a Dockefile is discouraged because it results in a different outcome every time it is executed. It's better to rely on an up-to-date base image.

Dockerfile
squatica: Ooof, I see you've been really busy lately. Let me catch up.

> Does the owner need to be changed? How does that translate across the mounted volume?

The user inside the container is root, but apache2-foreground automatically forks and then runs under www-data (this user is created during apache installation) so that's why we also have to run the cliupdate.php under www-data otherwise the first call to cliupdate would create the sqlite database with root permissions and then www-data would not be able to write in it.

>  What if www-data (or user with the same uid) does not exist outside of Docker container?

It doesn't have to exist outside the docker container. As I mentioned in another comment, this image was meant to be the production type with sources from git, and in such case we will never need to match the user inside with the user outside. Outside user will not need to write into these volumes (if so, they can do it by docker exec into this container and then they have full rights). The volume doesn't even need to be mounted to a directory in the host filesystem, docker will manage it somewhere in the background in /var/lib/docker or so.

squatica: > Do we need to create the config file when it does not exists? selfoss will happily read the data from defaults.ini when some keys or even the whole config.ini is missing.

This was actually something I had to work around - I had to create a separate directory for config and then symlink the config.ini into the right place. Because volumes can only be directories, not single files. Usually it's better to have user-changeable things such as configs in one directory, separate from the source files. And as mentioned previously, I'm checking for the existence of config.ini because we expect the volume to be reused, but we also want to initialize it when the volume is empty in the first run.

squatica: > Do we need to set the logger destination in the config, or might it make better sense forcing it through SELFOSS_LOGGER_DESTINATION environment variable.

I wanted to have the selfoss logs visible in `docker logs` of the whole container (and therefore also in `docker-compose logs` which is the first place to check if something doesn't work as expected). This is where docker includes the output of the CMD command, our CMD is apache2-foreground so `docker logs` will show access logs, but that's not so useful. In the meantime selfoss logs might contain important errors and they would be hidden un-intuitively in data/logs. By switching to `file:php://stderr` selfoss errors are included among the apache access logs. I'm not sure what SELFOSS_LOGGER_DESTINATION does - maybe you can achieve the same behavior with it?

squatica: > The while loop inside subprocess is very ugly.

It's actually quite normal thing in docker containers. Running two main processes (cron + apache) is discouraged - each container should only run one process. Same problem with systemd (plus a bunch of other permission related problems). While loop is simple and efficient and it will die when the container dies, so no problem. If we wanted to be completely pure about it, we would have to include something like supervisorctl and run apache and this cliupdate loop under supervisor. But that's too much hassle just for two processes. Alternative is to call the cliupdate from outside by `docker exec <container> php /var/www/html/cliupdate.php` but then we would force the user to do some extra setup outside of just docker-compose up.

Dockerfile
squatica: In multi staged builds the layers are not included in the final image - see [here](https://docs.docker.com/develop/develop-images/multistage-build/) - only the one file that's copied by `COPY --from=<previousStage>`

README.md
squatica: `npm run dev` and other dev steps should be used together with the dev docker image - I was planning to submit it once I have it cleaned up, you guys are too fast :-D

For the dev image we will just need to install php, apache, composer and npm all together in one image and then mount the source directory with a volume - the image itself will not contain any source code from selfoss, only all the tools that devs might use. I just didn't have enough time to test it, and installing npm together with php is a pain...

Dockerfile
squatica: About the entrypoint.sh - yes, thanks for putting it into utils, the only reason why I used echo in Dockerfile was to have two Dockerfiles - one for 2.18 and one for master and I didn't want to have two entrypoint.sh files (they differ in the chown lines).

Dockerfile
jtojnar: I would try

```patch
--- a/assets/package.json
+++ b/assets/package.json
@@ -1,4 +1,5 @@
 {
+  "name": "selfoss-assets",
   "dependencies": {
     "@fancyapps/fancybox": "^3.1.20",
     "@fortawesome/fontawesome-svg-core": "^1.2.25",

```

but it is weird. According to [this code](https://github.com/npm/cli/blob/44ddd0b528732028ea4f38ef15b5826f3f4e2ec1/lib/install/validate-args.js#L24-L32), it should ask for name first. Unless it comes from the top-level `package.json`.

What do you get by running `npm run install-dependencies > output.txt 2>&1`? I get [output.txt](https://github.com/SSilence/selfoss/files/4004269/output.txt).

akash07k: > I still get the same error, tried reinstalling node but no luck. :(
> don't know what's the root of this issue.
> Github is not allowing me to upload the output file so below's the pastebin URL:
> https://pastebin.com/3cJ46piE



> I would try
> 
> ```diff
> --- a/assets/package.json
> +++ b/assets/package.json
> @@ -1,4 +1,5 @@
>  {
> +  "name": "selfoss-assets",
>    "dependencies": {
>      "@fancyapps/fancybox": "^3.1.20",
>      "@fortawesome/fontawesome-svg-core": "^1.2.25",
> ```
> 
> but it is weird. According to [this code](https://github.com/npm/cli/blob/44ddd0b528732028ea4f38ef15b5826f3f4e2ec1/lib/install/validate-args.js#L24-L32), it should ask for name first. Unless it comes from the top-level `package.json`.
> 
> What do you get by running `npm run install-dependencies > output.txt 2>&1`? I get [output.txt](https://github.com/SSilence/selfoss/files/4004269/output.txt).

jtojnar: Weird. At this point I would pin it as some npm on Windows idiosyncrasy. Does running the command `npm install --production=false --prefix assets/` directly also fail? What about `cd assets && npm install --production=false`?

.dockerignore
akash07k: > Hmm, right, but I did it this way because many of the times we want to apply our local repo changes in the docker image itself. And by looking at this use case, it seemed the best choice to me. :-)
But you are absolutely right and then I think that targeting multiple use-cases will be a lot complex. :(
> We need to make a clear distinction about the purpose of this docker image. Two main types are:
> 
> * production
> * development
> 
> Where the production can further be devided into:
> 
> * sources come from git clone
> * sources come from the current directory (`ADD . /src` in Dockerfile)
> 
> My Dockerfile was intended as the production image with sources from git clone. Here is why: it can then be in a separate git repo (such as github.com/SSilence/selfoss-docker) and can be built completely separately yet always resulting in the same source files. It can also be built for some older tag, which is good either for comparison or when master is broken. Using production image with `ADD . /src` is problematic because then even your uncommitted files will be included in the image, which is a problem because you may accidentally push into docker hub something that doesn't exist anywhere in git - this is also how viruses get into docker images. In practice this approach is only used when the build is automated and done in a pipeline on the build server, where we can make sure it checks out a clean workspace from git before the docker build.
> 
> So if we keep this image as the production image with sources from git, we don't even need any .dockerignore file at all, because the build will run in a different directory - different git repo.

Dockerfile
akash07k: > Seems the right approach to me. @jtojnar opinions?

> > Do we need to set the logger destination in the config, or might it make better sense forcing it through SELFOSS_LOGGER_DESTINATION environment variable.
> 
> I wanted to have the selfoss logs visible in `docker logs` of the whole container (and therefore also in `docker-compose logs` which is the first place to check if something doesn't work as expected). This is where docker includes the output of the CMD command, our CMD is apache2-foreground so `docker logs` will show access logs, but that's not so useful. In the meantime selfoss logs might contain important errors and they would be hidden un-intuitively in data/logs. By switching to `file:php://stderr` selfoss errors are included among the apache access logs. I'm not sure what SELFOSS_LOGGER_DESTINATION does - maybe you can achieve the same behavior with it?

akash07k: > can't we use cron? will it cause issue?
> > The while loop inside subprocess is very ugly.
> 
> It's actually quite normal thing in docker containers. Running two main processes (cron + apache) is discouraged - each container should only run one process. Same problem with systemd (plus a bunch of other permission related problems). While loop is simple and efficient and it will die when the container dies, so no problem. If we wanted to be completely pure about it, we would have to include something like supervisorctl and run apache and this cliupdate loop under supervisor. But that's too much hassle just for two processes. Alternative is to call the cliupdate from outside by `docker exec <container> php /var/www/html/cliupdate.php` but then we would force the user to do some extra setup outside of just docker-compose up.

README.md
akash07k: > I can't figure out that why isn't it taking my config.ini into account then. may be some mouting/linking issue?
> The Dockerfile specifies the VOLUME for the config directory, so the directory will already be created inside. Then in the entrypoint we check if config.ini is present and create it if not.
> 
> So this can be used with or without docker-compose.
> 
> Running docker run directly even without the -v option will still create the volume, only it will not be named, but docker will still remember which volume belongs to which container and will use it next time you run the same container. Using docker-compose only allows you to have more flexibility with the volumes, such as you can easily create another container that uses the same volume and runs a backup on it, etc. Or you can easily join this container and its volumes with another containers in a swarm (which is why I was even looking into selfoss in the first place - I wanted to add an rss aggregator container into my existing project with many other containers)

Dockerfile
akash07k: > In multi staged builds the layers are not included in the final image - see [here](https://docs.docker.com/develop/develop-images/multistage-build/) - only the one file that's copied by `COPY --from=<previousStage>`

Oh, that's nice.
I'll remove the apt-get upgrade then

README.md
akash07k: > I'm actually not that experienced in docker.
>  Now I too feel that I did it so early.
> You seem to have in depth knowledge of docker, and if you are willing to take it completely, then can I remove this pull request entirely?
> And I feel that I'll be able to learn a lot from you. :-)
> `npm run dev` and other dev steps should be used together with the dev docker image - I was planning to submit it once I have it cleaned up, you guys are too fast :-D
> 
> For the dev image we will just need to install php, apache, composer and npm all together in one image and then mount the source directory with a volume - the image itself will not contain any source code from selfoss, only all the tools that devs might use. I just didn't have enough time to test it, and installing npm together with php is a pain...

Dockerfile
akash07k: > hmm, seems valid point. I'll revert the changes.

> I like to use DEBIAN_FRONTEND=noninteractive directly on the line before apt-get install, because if it's configured like this globally at the beginning, it will remain set even when the container is running, so even if you later `docker exec` into the container and want to `apt-get install` something, apt will think you are still non-interactive.

Dockerfile
akash07k: > will remove it. :-)

> apt-get upgrade in a Dockefile is discouraged because it results in a different outcome every time it is executed. It's better to rely on an up-to-date base image.

Dockerfile
akash07k: > I'll check with these commands too.
> but it is not related to Windows because same error comes while in the docker container too and on my linux server as well.
I think that it is related to node version? which version you are using?


> Weird. At this point I would pin it as some npm on Windows idiosyncrasy. Does running the command `npm install --production=false --prefix assets/` directly also fail? What about `cd assets && npm install --production=false`?

jtojnar: I tried node 10.17.0 with npm 6.13.0, node 10.18.0 with npm 6.13.4, and node 13.3.0 with 6.13.1.

akash07k: I think that I should try it with node version 13.3.0 then, may be it will work. 

> I tried node 10.17.0 with npm 6.13.0, node 10.18.0 with npm 6.13.4, and node 13.3.0 with 6.13.1.

Dockerfile
```diff
@@ -0,0 +1,42 @@
+FROM ubuntu:bionic as source
+RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install --no-install-recommends -y && \
+    rm -rf /var/lib/apt/lists/*
+
+COPY . selfoss/
+
```
jtojnar: I would remove  this whole part and use `COPY . selfoss/` in the `composer` stage.

akash07k: > Hmm, seems good. going to update after testing whether it breaks something or not.

> I would remove this whole part and use `COPY . selfoss/` in the `composer` stage.

akash07k: > I would remove this whole part and use `COPY . selfoss/` in the `composer` stage.

It will reduce the build timing as well.

README.md
squatica: Depends. We could transform this PR into the dev docker image, and then have another one for the production image. I'll send you the Dockerfile soon once I finish dealing with node user, grrr.. (can I create a PR against your PR? I've never done that before...) But we definitely want to keep the entrypoint in utils because even the production image will depend on it.

README.md
squatica: @akash07k Can you join the gitter room? I can help you debug the config issue, chatroom is a better tool for that. I'll be online for the next hour or two

README.md
akash07k: > seems the good idea.
> sure, you can create the pull request against this. Can we do that? I don't know that how to acomplish it or whether it is doable or not.
:-)
> Depends. We could transform this PR into the dev docker image, and then have another one for the production image. I'll send you the Dockerfile soon once I finish dealing with node user, grrr.. (can I create a PR against your PR? I've never done that before...) But we definitely want to keep the entrypoint in utils because even the production image will depend on it.

README.md
akash07k: > I'm sorry, I just came online and checked your comment. are you still online? and I haven't joined any gitter chatroom untill now, is there any link to join?
> Please excuse me for this but I have only heard about gitter but never used it. Though, I'm keen to try it. :-)

> @akash07k Can you join the gitter room? I can help you debug the config issue, chatroom is a better tool for that. I'll be online for the next hour or two

squatica: yes I'm here now: https://gitter.im/fossar/selfoss you will access it with your github account, just confirm the permissions

akash07k:
Ok. loging in  :-)

README.md
jtojnar: Alternately, you can use [Matrix/Riot](https://matrix.to/#/#selfoss:matrix.org), or [IRC](https://irc.gitter.im/) to connect to the room, if you prefer.

akash07k:
Oh wow! seems interesting options. going to check them out too.
Thanks a lot bro :-)

PR #1164, opened 2019-12-29
Decouple classes using dependency injection

jtojnar:
Instead of instantiating its dependencies in each class, we are going to switch to dependency injection. We will declare the dependencies using typehinted constructor arguments and pass already instantiated dependency objects to the constructor. This will loosen the coupling between classes and allow us to pass mock objects for testing.

Rather than creating and passing instances of dependencies manually, we will let Dice dependency injection container library instantiate them for us. We are setting `CONTAINER` value in fatfree’s Hive to a function wrapping around Dice->create method (Dice does not support PSR-11), which will make fatfree ask Dice to instantiate controllers. Dice will recursively instantiate all of the controller’s dependencies and then instantiate the controller itself.

At the moment, only a small fraction of selfoss was converted but more is coming soon.


jtojnar:
The hardest part to convert are the DAOs, which are pretty messy. They use something like [bridge design pattern](https://en.wikipedia.org/wiki/Bridge_pattern) but everything is spiked with global data container/service locator (f3’s Hive). Plus inheritance being used as code deduplication technique.

Class Diagram1

Ideally, we would end up with something like the following:

Class Diagram4

Bridges will need to take dependency container and create the backends using that.

jtojnar:
[Late static binding](https://www.php.net/manual/en/language.oop5.late-static-bindings.php) is pretty awesome:

```php
<?php
class FooStmt {
	public static function test() {
		echo "FooStmt::test" . PHP_EOL;
	}
}

class BarStmt extends FooStmt {
	public static function test() {
		echo "BarStmt::test" . PHP_EOL;
	}
}

class Foo {
	static $stmt = FooStmt::class;

	public static function test() {
		echo get_called_class() . PHP_EOL;
		echo self::$stmt . PHP_EOL;
		echo static::$stmt . PHP_EOL;

		static::$stmt::test();
	}
}

class Bar extends Foo {
	static $stmt = BarStmt::class;
}

Foo::test();
Bar::test();
```

produces

```
Foo
FooStmt
FooStmt
FooStmt::test
Bar
FooStmt
BarStmt
BarStmt::test
```

We can use that to avoid inheriting from `Database`.

jtojnar:
With late static bindings it is closer to

Class Diagram3

PR #1170, opened 2020-02-07
Docker

squatica:
Hi,
building on top of the PR #1162 updated the branches with current master and joined mine and @akash07k 's changes together into this PR. I suggest to squash all the commits together (maybe except deasync and '--dev'), because it's pretty confusing as it is. I left them separate for now just in case you prefer to look at all the single commits.


squatica:
@jtojnar What do you think? Can it be merged? Or anything else necessary?

.dockerignore
```diff
@@ -0,0 +1,16 @@
+/assets/.cache
```
jtojnar: Is there not a way to make the `.dockerignore` copy `.gitignore`? Keeping the two in sync will be annoying.

.env.dist
```diff
@@ -0,0 +1,2 @@
+UID=1000
```
jtojnar: Is there not a better way to mount the dev directory without needing uid/gid? Not only this adds to the clutter but it also feels a bit wonky.

README.md
```diff
@@ -77,6 +77,36 @@ If you want to create a package with all the dependencies bundled, you can run `
 
 Every patch is expected to adhere to our coding style, which is checked automatically by Travis. You can install the checkers locally using `npm run install-dependencies`, and then run the checks using `npm run check` before submitting a pull request. There is also `npm run fix`, that will attempt to fix the formatting.
 
+## Dockerizing selfoss
```
jtojnar: The readme is also too cluttered. I would move this section to `docs/docker.md` and link it from the `README`.

assets/package-lock.json
```diff
@@ -1888,10 +1888,13 @@
       "dev": true
```
jtojnar: This should not be part of this PR. Will cherry-pick to master.

package.json
```diff
@@ -26,7 +26,7 @@
     "fix:server": "composer run-script fix",
     "install-dependencies": "npm run install-dependencies:client && npm run install-dependencies:server",
     "install-dependencies:client": "npm install --production=false --prefix assets/",
-    "install-dependencies:server": "composer install --dev",
```
jtojnar: Same here.

utils/docker/entrypoint.sh
```diff
@@ -0,0 +1,22 @@
...
+    /var/www/html/data/cache \
+    /var/www/html/data/favicons \
+    /var/www/html/data/logs \
+    /var/www/html/data/thumbnails \
+    /var/www/html/data/sqlite
+
+# Create a config file when one does not exist
+if [[ ! -f config/config.ini ]]; then
+    cp defaults.ini config/config.ini \
+    && sed -i 's#^logger_destination=.*#logger_destination=file:php://stderr#' config/config.ini
```
jtojnar:
```suggestion
    sed 's#^logger_destination=.*#logger_destination=file:php://stderr#' defaults.ini > config/config.ini
```

.dockerignore
squatica: They do two different things, they don't necessarily need to be the same. Dockerignore prevents files from being used during build, which speeds up the build and makes smaller layers. But honestly, if we always build the official image from a clean repository, we don't even need a dockerignore file.

utils/docker/entrypoint.sh
squatica: Good catch, thanks.

.env.dist
squatica: Unfortunately, this is a years-long issue with docker and volumes - https://github.com/moby/moby/issues/2259. And in this case we need the executor of `npm run dev` to match the owner of the mounted directory, which can be anything. This solutions is one of the least complicated and working.

README.md
squatica: I moved it to utils/docker, so that it gets displayed to anyone lurking in that directory.

squatica:
OK, I tested how it works with docker-compose files in utils/docker and it seems to be fine. It's not standard but works.

About Apache - I'd rather keep it, because there are a bunch of .htaccess files which will all stop working under php builtin server. Also php builtin server should not be used for production. What do you find complex about the Dockerfiles? They seem quite small to me, I've seen much worse :-D The production one is split into three parts using multi-staged build because we don't need npm nor composer for the production runtime environment. And the dev dockerfile basically only installs node and composer according to their official installation instructions. One minor simplification will become possible once the config.ini location is configurable or moved into data directory as suggested by #913 .

akash07k:
Thanks a lot @squatica for this.
It's working great so far.
Actually I was out of station so couldn't pay attention to it till now.
Thanks a tons for your efforts into this!

squatica:
Hey @akash07k nice to hear from you again, I was worried what happened to you :smiley: 

@jtojnar I created a docker hub account for selfoss, I'm going to configure automatic builds of the production image. Let me know how I can transfer the ownership to you.

akash07k:
@squatica 
Yes buddy. actually I went on vacations for sometime and my laptop wasn't with me.
I'm back now.
how are you?
and I'm very greatful for you for your efforts.
I pinged you on gitter a month back but probably you were busy then.

> Hey @akash07k nice to hear from you again, I was worried what happened to you 😃
> 
> @jtojnar I created a docker hub account for selfoss, I'm going to configure automatic builds of the production image. Let me know how I can transfer the ownership to you.

akash07k:
Isn't it yet ready for merge?

jtojnar:
The main blocker is that we do not have anyone knowledgable in Docker to properly review this.

akash07k:
:(

kolaente:
Looks quite good, I have a few ideas:

1. I would suggest creating the dev image on top of the production image - that way you get the benefit of having the same docker image for dev and prod (well, almost at least).
2. Another approach would be to have a seperate "workspace" container which contains all the tools needed for development (composer etc) and has the same mounts as the prod image. Both would run side by side, but the prod image based container would only host the files - and be the same as any "real" prod deployment.

akash07k:
> Looks quite good, I have a few ideas:
> 
> 1. I would suggest creating the dev image on top of the production image - that way you get the benefit of having the same docker image for dev and prod (well, almost at least).
> 2. Another approach would be to have a seperate "workspace" container which contains all the tools needed for development (composer etc) and has the same mounts as the prod image. Both would run side by side, but the prod image based container would only host the files - and be the same as any "real" prod deployment. This one would then also do things like running the frontend watcher on startup.
Good ideas,
I'm not that expert in Docker otherwise I would have tried working on these.
@squatica opinions bro?

akash07k:
It seems that now there are some package conflicts in package.lock.json due to the dependency updates. :-)

jtojnar:
Sorry, I do not really use docker so it has been difficult for me to allocate time to it. I will try to do something with #1271 this week.

Issue #1185, opened 2020-05-03
FullText RSS issue with php links containing arguments


sonicnkt:
Hi,
Im not sure if this is a issue with selfoss itself or the graby library but i have an issue with scraping webpages that have php links to the articles like the following: 
`https://webpage.com/news.php?item=news&article=test`

For some reason the page that will be scraped is the base php page `https://webpage.com/news.php` and the arguments are not really used for processing. 

One example webpage would be phoronix.com:
`https://www.phoronix.com/scan.php?page=news_item&px=Steam-Survey-April-2020`

looking at the debug logs shows that graby tries to scrape the following link where it replaces the & symbol with the xml escape `&amp`;:
`https://www.phoronix.com/scan.php?page=news_item&amp;px=Steam-Survey-April-2020`

Im not sure if this "escaped" version is just displayed in the logs or if this is the url which gets processed but if it is that could be the issues im facing as trying to open this escaped link in a browser has the same results as im experiencing. 

Like i said im not sure if this is a issue with selfoss or graby but i have to start investigating somwhere :) 

PS: im currently not running the latest 2.19 Snapshots as i could not get this running on my system (Linux, NGINX), my current build is a few month older. I am really sorry if this issues was fixed allready.

jtojnar:
Thanks for reporting. Can confirm it still happens on master. Apparently, SimplePie htmlencodes the links :woman_facepalming::

1. https://github.com/SSilence/selfoss/blob/76452b0392831c6dd6249dd00c48eaaaf4da74e3/src/spouts/rss/fulltextrss.php#L58
2. https://github.com/simplepie/simplepie/blob/ae49e2201b6da9c808e5dac437aca356a11831b4/library/SimplePie/Item.php#L960
3. https://github.com/simplepie/simplepie/blob/ae49e2201b6da9c808e5dac437aca356a11831b4/library/SimplePie/Item.php#L1010
4. https://github.com/simplepie/simplepie/blob/ae49e2201b6da9c808e5dac437aca356a11831b4/library/SimplePie/Sanitize.php#L390-L393

jtojnar:
Also please open an issue for any troubles you had when installing. I would like to release 2.19 soon so pointing out any bugs or omissions in documentation would be very appreciated.

sonicnkt:
ah cool you could identify the issue so quickly! 
I will take a look again at a newer snapshot again and report back.

sonicnkt:
Update:
[...]

PR #1187, opened 2020-05-03
Do not escape links

jtojnar:
Fixes: https://github.com/SSilence/selfoss/issues/1185


jtojnar:
Replaced by: https://github.com/SSilence/selfoss/pull/1188

PR #1190, opened 2020-05-04
Refactor image fetching

jtojnar:
Avoid repeating requests and clean up the old spaghetti.

### Customization changes

To allow distinguishing between an icon coming from the feed item and one from the feed itself, we added a new optional `getSourceIcon()` method. When the `getIcon()` method returns `null` for an item or the item icon cannot be fetched, the source icon will be used as a fallback.

Both `getSourceIcon()` and `getIcon()` can return `null` now when an icon is not available.

### DIFF
--- a/src/spouts/rss/feed.php
+++ b/src/spouts/rss/feed.php
@@ -36,9 +36,6 @@ class feed extends \spouts\spout {
     /** @var ?string URL of the source */
     protected $htmlUrl = null;
 
-    /** @var ?string URL of the favicon */
-    protected $faviconUrl = null;
-
     /** @var Logger */
     private $logger;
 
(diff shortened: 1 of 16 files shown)


src/spouts/rss/feed.php
```diff
@@ -103,35 +100,42 @@ public function getContent() {
...
+        if ($feedLogoUrl && ($iconData = $this->imageHelper->fetchFavicon($feedLogoUrl)) !== null) {
+            list($faviconUrl, $iconBlob) = $iconData;
 
-            return $this->faviconUrl;
+            $aspectRatio = $iconBlob->getWidth() / $iconBlob->getHeight();
+
+            if (0.8 < $aspectRatio && $aspectRatio < 1.3) {
+                $this->logger->debug('icon: using feed logo: ' . $faviconUrl);
+
+                return $faviconUrl;
```
jtojnar, line 117: Eventually, we will want to return the blob as well so that it does not have to be downloaded again but we have already decreased the number of downloads a lot and I do not want to delay this any longer.

PR #1205, opened 2020-07-14
Support search using regex

jtojnar:
When user searches a term starting and ending with a forward slash, the items will be matched against the regular expression between the slashes.

The syntax of the regular expressions is database-engine specific:

* MySQL uses POSIX extended https://dev.mysql.com/doc/refman/5.7/en/regexp.html
* PostgreSQL uses POSIX extended with some extensions https://www.sqlite.org/lang_expr.html#the_like_glob_regexp_and_match_operators
* SQLite uses PHP’s PCRE https://www.postgresql.org/docs/12/functions-matching.html#FUNCTIONS-POSIX-REGEXP, https://www.php.net/manual/en/book.pcre.php

------

~~The filtering works but the interface is still incomplete. In particular, the matches will not be highlighted in the articles, and the search regex will be shown split on spaces.~~ Fixed.


phicha20224 (approved):

jtojnar:
I want to base this onto #1216, which will make it easier to distinguish the search terms in the UI.

PR #1219, opened 2020-09-14
Website revamp

jtojnar:
Some progress on #979, needed for the website to accommodate more documentation.

I have switched from a single HTML page to a semi-structured set of CommonMark documents built using a static site generator. I have chosen [Zola](https://www.getzola.org/) since it seemed to require the least amount of boilerplate (compared to Hexo, Hugo and Hakyll).

I am not a designer so the website does not look that good at the moment, will try to improve it later.

Also added [Nix](https://nixos.org/nix/) support, which I have been using locally for last few years, to make obtaining dependencies (npm, composer, phpstan, zola) more convenient.


jtojnar:
@SSilence ideally, we would add a CI script for auto-deployment to the website but that would require either adding credentials to the pipeline or setting up GitHub pages and pointing a `CNAME` record there. The latter would be cleaner but would break the discussion forum.

SSilence:
@jtojnar: I think a new github.io page would be the easiest solution. I can set a permanent redirect from selfoss.aditu.de. We can also make a new github organization and move the selfoss repository. Then its easier to set access rights and I can give you admin rights for further configuration (ci script for the page generation).

jtojnar:
@SSilence I am wary of using domains we do not control, I have seen far too many (not only free) services to close. We could still use the GitHub pages as a web host and [point the domain to that](https://docs.github.com/en/github/working-with-github-pages/configuring-a-custom-domain-for-your-github-pages-site). That would allow us to transparently switch when GitHub eventually goes down.

The forum could be converted to HTML and archived and people pointed to GitHub for new discussions, which is already happening. It is little sad that we depend on proprietary service for issues but IMHO convenience wins here.

But the website business can wait, I would like to release 2.19 in October.

I made the [fossar organisation](https://github.com/fossar) for this, since [selfoss](https://github.com/selfoss) name was already taken. (It means “waterfalls” in Icelandic.)

SSilence:
@jtojnar I transfered the repository into the fossar organisation. Not bad, the old url redirects to new repo url. I can register a dedicated domain for selfoss. e.g. selfoss.org with CNAME for the github.io page. You are right, no one knows what happens in future with github.io domains. Thanks for your commitment with this project!

jtojnar:
@SSilence Thanks, I have updated the docs in this PR with the new URL.

Dedicated domain would be nice but choosing one is a pain I would like to defer for as long as possible. `selfoss.org` is already registered by a domain squatter or something, the [affair around `.io` TLD](https://en.wikipedia.org/wiki/.io#Controversy) is something I would rather avoid, and the other schemes like `getselfoss.com` go in and out of fashion every year.

Issue #1220, opened 2020-09-21
Formatted text


davidoskky:
Using fulltextrss the full html of the article should be recovered.
Would it be possible to maintain the formatting of the original text?
What I would like is to have italic and bold sections as such in selfoss. Sometimes in articles the presence of a sentence in italic or a bold word may translate in a significant difference in the meaning.

jtojnar:
I would expect the formatting to be preserved, unless it is done in CSS, which might be hard. Do you have an example of site where such formatting occurs?

davidoskky:
I did not find a recent one in english...
https://ilblogdellasci.wordpress.com/feed/
In the last article (21/09/2020) there is a large paragraph in italic, this is not shown inside selfoss.

jtojnar:
Hmm, looks like the css reset we use is eating the style for `<em>`. For now, you can put the following into `user.css` in your selfoss directory:


```css
.entry-content i, .entry-content em {
    font-style: italic;
}

.entry-content b, .entry-content strong {
    font-weight: bold;
}
```

PR #1221, opened 2020-09-21
client: re-add styles to content

jtojnar:
Reset.css removes all styling including from elements like `<em>` or `<strong>`. Let’s reapply it in entry contents.

I had to switch to SASS to be able to use scoped @import.

The `unreset` library contains stuff we do not need (semantic layout elements) but it is still nicer than writing it from scratch.

Alternately, we could switch to something like normalize.css and remove stuff we do not want (e.g. list styling for `<ul>` in the sidebar) ourselves but that is more work and I am not sure how much sense it makes for us. https://css-tricks.com/reboot-resets-reasoning/

Fixes: #1220


niol:
I do not understand why we import some `unreset` stuff instead of styling `em` etc. elements like you suggested in the bug report.

jtojnar:
The reset.css removes styling for all elements so we would need to add most of https://github.com/ixkaito/unreset.css/blob/master/_unreset.scss.

niol (approved):

PR #1232, opened 2020-11-17
Article reading time

davidoskky:
This addition displays an estimated reading time near the title of each article.
It allows me to estimate how long an article is before opening it.
It is just a small hack I made for my own use and I do not know if it actually is an useful feature for others... if it's not feel free to delete this PR.

### DIFF
--- a/assets/locale/en.json
+++ b/assets/locale/en.json
@@ -99,5 +99,6 @@
   "lang_error_share_native": "Unable to share item.",
   "lang_info_url_copied": "URL copied to clipboard.",
   "lang_close_entry": "Close",
-  "lang_article_actions": "Actions"
+  "lang_article_actions": "Actions",
+  "lang_article_reading_time": "{0} min read"
 }
--- a/src/templates/item.phtml
+++ b/src/templates/item.phtml
@@ -57,6 +57,12 @@
     <!-- datetime -->
     <a href="<?= \helpers\Anonymizer::anonymize($this->item['link']); ?>" class="entry-datetime" target="_blank" rel="noopener noreferrer"></a>
 
+    <!-- read time -->
+    <?php if (\F3::get('reading_speed_wpm') > 0) : ?>
+    <span class="entry-separator">&bull;</span>
+    <span class="entry-readtime"><?= \F3::get('lang_article_reading_time', round((str_word_count(strip_tags($content)) / \F3::get('reading_speed_wpm'))))?></span>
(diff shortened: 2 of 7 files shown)


jtojnar:
Thanks. This is useful feature, my only worry is that it will crowd the interface even more. Maybe we could default the config value to 0 and only show the time when user explicitly sets their reading speed?

davidoskky:
Indeed it could become crowded; I edited the PR, this way the feature is disabled by default.

assets/locale/en.json
```diff
@@ -99,5 +99,6 @@
   "lang_error_share_native": "Unable to share item.",
   "lang_info_url_copied": "URL copied to clipboard.",
   "lang_close_entry": "Close",
-  "lang_article_actions": "Actions"
+  "lang_article_actions": "Actions",
+  "lang_article_minutes": "min"
```
jtojnar, line 103: It would be nice to use so that translators have more control.

```suggestion
  "lang_article_minutes": "{0} min"
```

src/templates/item.phtml
```diff
@@ -57,6 +57,12 @@
     <!-- datetime -->
     <a href="<?= \helpers\Anonymizer::anonymize($this->item['link']); ?>" class="entry-datetime" target="_blank" rel="noopener noreferrer"></a>
 
+    <!-- read time -->
+    <?php if (\F3::get('reading_speed_wpm') > 0) : ?>
+    <span class="entry-separator">&bull;</span>
+    <span class="entry-readtime"><?php echo round((str_word_count(strip_tags($content)) / \F3::get('reading_speed_wpm'))); ?> <?= \F3::get('lang_article_minutes')?></span>
```
jtojnar, line 63: Then you should be able to use it as:

```suggestion
    <span class="entry-readtime"><?= \F3::get('lang_article_minutes', round((str_word_count(strip_tags($content)) / \F3::get('reading_speed_wpm'))))?></span>
```

jtojnar (commented):
Looks good.

Could you also add description of the option to https://github.com/fossar/selfoss/blob/master/docs/content/docs/administration/options.md? If not, I can do that before merging.

And just to be sure, are you okay with licensing your contribution as GNU GPL 3 and any later version?

davidoskky:
I hope the description I added is clear enough, feel free to edit that if it is not.
I am fine with GNU GPL 3 and later versions.

jtojnar:
Thanks.

PR #1233, opened 2020-12-11
Create basic integration testing

jtojnar:
To be able to see if PHP 8 works.

Currently fails due to `requests` library seemingly not to being able to extract cookies.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,8 @@
 # variables need to be set.
 language: php
 
+dist: xenial
+
 matrix:
     include:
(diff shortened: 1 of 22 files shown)


.travis.yml
```diff
@@ -29,6 +32,9 @@ cache:
 
 before_install:
     - nvm install 12
+    - pyenv install 3.6.3
```
jtojnar, line 35: This is rather slow, eventually we will want to switch to Nix for all our dependency related needs.

PR #1260, opened 2021-04-26
spouts/YouTube: add support for playlists

jtojnar:


jtojnar:
Turns out anonymous classes are not supported on PHP 5.6 so we either need to re-engineer that or wait until after we release selfoss 2.19.

jtojnar:
Stubbing works with PHP 5 and is even shorter.

toonn:
🎉 Thanks so much!

PR #1261, opened 2021-04-27
Do not use F3 for configuration management

jtojnar:
Fatfree is badly designed and should not have lived past 2005. Most notably, it relies on a single huge singleton object, which makes testing difficult. What is worse, the unholy things it does to error handling makes PHPStan and PHPUnit crash. For these reasons and other pains, selfoss will be gradually removing its use. Configuration values are read throughout the codebase so I decided to tackle those first.

Ideally, we would construct a configuration container using something like <https://gitlab.com/Khartir/typed-config> but that requires more recent PHP version than we support. Other configuration libraries either suffer from the same problem, are untyped like f3 was (domains of the options are known so we should not have to cast the string loaded from the ini file at the point of use), or are completely undocumented. Custom solution had to be implemented.

This patch adds `helpers\Configuration` class, which will load data from `config.ini` and environment variables and store them in properties. The class is injected by Dice into whichever other class that needs access to configuration. In addition to reducing the percentage of code paths tainted by F3,
it will also bring better developer experience since editors and other tools can make use of the property annotations.

Since the default values of configuration option are now stored as default values of the class properties, `defaults.ini` is no longer used. It will still be included in the distribution zipball for convenience (generated from the code.

In the future, we can move the option descriptions into the class and have the options documentation generated as well.


desbest:
Out of curiosity, what is wrong with Fat Free Framework? Why don't you recommend it?

jtojnar:
There is just so many issues and problems with it I had to work around in selfoss. Many of them just stem out of its old age and are impossible to fix without breaking backwards compatibility and massive rewrite unfortunately.

These are just a few I remember at the top of my head:

1. It is monolithic – ­you need to pull in the whole framework, even if you do not want to use some of its components. And what is worse, you cannot avoid using some components like [error handler](https://github.com/bcosca/fatfree/issues/999), and the default one is [extremely invasive](https://github.com/fossar/selfoss/blob/7bafa84f3b992f293c1f1f53eb4c3ca25e6598df/src/common.php#L21-L25).
  - It is curious that the framework calls itself Fatfree when there is so much code I would cut out. One could take great inspiration from how [Nette framework split](https://github.com/nette/nette/releases/tag/v2.2.0) and also [Symfony split](https://symfony.com/blog/symfony2-components-as-standalone-packages).
2. It has questionable design:
  - It uses a singleton class that works as a configuration container. Well, you can create a new instance of the `Base` class but it will still internally [work with the singleton](https://github.com/bcosca/fatfree-core/blob/b35fe7ae88f6f3667c50c36563b50e5a1ec36bb1/base.php#L422) behind your back. That makes testing hard since you cannot just inject a fresh F3 instance into a test.
  - The container stores random untyped values so you have no benefit of static checking.
  - The `Base` class modifies many PHP configuration values (error handler) on its construction. And what is worse, the file containing the class [automatically creates the `Base` class instance](https://github.com/bcosca/fatfree-core/blob/b35fe7ae88f6f3667c50c36563b50e5a1ec36bb1/base.php#L3414). This means that when you even mention `F3` and autoloader includes it, your PHP state changes (breaking error handlers PHPUnit set up, for example). ← This was the last straw for me.
3. The [code base](https://github.com/bcosca/fatfree-core/blob/b35fe7ae88f6f3667c50c36563b50e5a1ec36bb1/base.php) is messy and unorganized – just look at it.
  - The code trades clarity for terseness (e.g. lack of brackets for control structures, complex predicates in `if` statement conditions, assigning variables deep within those…). That makes debugging unnecessarily harder.
  - It also makes it much easier to introduce bugs.
4. It [does not use namespaces](https://github.com/bcosca/fatfree-core/issues/110) so we cannot use some class names or we get name clashes.
  - Okay, using namespaces in selfoss that would allow us to have `Image` class. But sometimes I want to test an old library (I would add namespace support to it when I decide to adopt them) and it clashes too.

jtojnar:
Fuck, I forgot we still use this piece of work. Just spent almost three hours debugging why Symfony console in #1283  fails with `Command "/" is not defined.`. Turns out Fatfree adds `/` to `$_SERVER['argv']` when no arguments are provided 🤦‍♀️

PR #1283, opened 2021-11-03
Add a selfoss CLI tool

jtojnar:
This introduces two commands for dumping and loading the database contents:

- `php bin/selfoss db:export path.json`
- `php bin/selfoss db:import path.json`

Those are useful for testing, as well as for moving data from one database backend to another.


jtojnar:
Currently, `datetime`s exported from SQLite cannot be imported to MySQL, we need some normalization.

PR #1312, opened 2022-01-23
Fix compatibility with PHP 8.1

jtojnar:
Fixes: https://github.com/fossar/selfoss/issues/1294


jtojnar:
Test complains

```
Deprecated: Return type of GuzzleHttp\Handler\MockHandler::count() should either be compatible with Countable::count(): int, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /home/runner/work/selfoss/selfoss/vendor/guzzlehttp/guzzle/src/Handler/MockHandler.php on line 173
```

but it does not seem to make it fail. Opened https://github.com/guzzle/guzzle/pull/2983.

We should also wait for new version of fatfree to be released.

PR #1315, opened 2022-01-24
composer: Switch to latest dependencies

jtojnar:
We now target PHP 8.1, compatibility with older PHP versions will be provided using code transformation with [Rector](https://getrector.org/).

- Remove config.platform.php from composer since it forces versions to be installed.
- Dice is now immutable, we need to overwrite the variable. Also change the container to inject the proper Dice instance.
- monolog\Logger expects a string argument so we need to set up DI to provide it one in unit tests.
- monolog deprecated the warn method.
- Graby now depends on tidy


jtojnar:
This was superseded by #1392

PR #1343, opened 2022-07-10
Allow overriding path to config file

jtojnar:
When `SELFOSS_CONFIG_PATH` environment variable is set, selfoss will load the referenced file instead of config.ini.


jtojnar:
Not sure if this feature is useful to anyone. I no longer need it.

Issue #1344, opened 2022-07-12
Accessibility of sources-list


heull001:
The source-list (selfoss.link/manage/sources) is not very accessible yet. Every function is usable, but its not comfortable to navigate with screenreader here. The first problem I think, form is not the right tag to group an item. I think, article or maybe section would be better here. By using form, not every screenreader notices that there should be a newline between the items.

Also it would be helpfull to tag the source-titel as headline, most screenreader-users use headlines to navigate.

jtojnar:
Thanks for reporting. I tried to improve it in https://github.com/fossar/selfoss/pull/1345. Article did not feel right so I went with list items. You can download the pre-built version at https://github.com/fossar/selfoss/actions/runs/2658340140.

jtojnar:
Good idea, pushed another revision. https://github.com/fossar/selfoss/runs/7324516538?check_suite_focus=true

I am still unable to test it well since the only screen reader I have access to is Orca and it seems to freeze my web browser.

jtojnar:
I use GNOME and Firefox freezes for me when enabling it. But it might just be that I have too many tabs open and too many sources in selfoss.

heull001:
That's possible. So, if you need someone to test something with a11y, 
especially for screenreader-users, just contact me, I'm using selfoss 
several times a day, so I'm very interested in good a11y.

akash07k:
I too can help regarding testing accessibility.
I'm a full time accessibility consultent and developer

jtojnar:
Thanks for the offer. I think feedback from the actual users is the most important for improving the usability. Particularly if there is any part of the UI that is inaccessible.

# Colour chooser accessibility

Actually, one change I am not sure about in 2.19 is making the tag colour chooser accessible via keyboard. The side effect is that the user now needs to press <kbd>Tab</kbd> twice to jump to the next tag in the list in the sidebar. That might be annoying to e.g. blind users who use keyboard to navigate but have no use for coloured tags. Or perhaps they use more sophisticated navigation tools than pressing <kbd>Tab</kbd> so it does not bother them at all.

On the other hand people hard of seeing might benefit from the ability to change tag colour that might not have been available to them previously if they are not able to use a mouse.

[...]

PR #1345, opened 2022-07-12
client: Improve screen reader navigability of sources

jtojnar:
- Turn sources into a list items rather than forms, only the inside is an actual form.
- Turn the source titles into headings.
- Mark the edit button as collapsed.
- Add a placeholder when no sources are present.

Fixes: #1344


akash07k:
Thanks @jtojnar 
Really good work.
I'm gonna try Selfoss today after the gap of long 2.5 years.

Issue #1357, opened 2022-09-28
/update specs


davidoskky:
Hi, I'm sorry about opening an issue just regarding some information about the api, but couldn't find it in the documentation.
What are the plausible response of /update?
I reckon that if everything works fine you get a successful response containing "finished"
Are there other possible response messages?

jtojnar:
That is completely fine. After all, the API endpoint is not documented, which is either a documentation bug, or an intentional omission.

Both `controllers\Sources\Update::updateAll` (`POST /update`) and `controllers\Sources\Update::update` (`POST /source/:sourceId/update`) actions only return `finished` when completed and [`unallowed access`](https://github.com/fossar/selfoss/blob/7b4b887730e0dd5a5b3457928f5c01f96f4a4362/src/controllers/Sources/Update.php#L32) when no user is not logged in and updating sources is not allowed for public. But the later also currently has 200 status code, which is suboptimal.

Ideally, we would also report back errors that occur during updating a single source, and update progress for updating all sources.

jtojnar:
#1428 will add alternative response format to `GET /update`.

PR #1360, opened 2022-09-30
Update API docs & deprecate some older ones

jtojnar:
- Emphasize Cookie auth in docs
- Replace `GET /logout` with `DELETE /api/session/current`
- Deprecate `POST /source/delete/:id`
- Make `POST /source/update` and `POST /source/:id/update` return 403 on authorization failure (cc @davidoskky for #1357)

cc @aminecmi

### DIFF
--- a/src/constants.php
+++ b/src/constants.php
@@ -7,4 +7,4 @@ const SELFOSS_VERSION = '2.19-SNAPSHOT';
 // independent of selfoss version
 // needs to be bumped each time public API is changed (follows semver)
 // keep in sync with docs/api-description.json
-const SELFOSS_API_VERSION = '4.1.0';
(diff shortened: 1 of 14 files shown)


src/constants.php
```diff
@@ -7,4 +7,4 @@
 // independent of selfoss version
 // needs to be bumped each time public API is changed (follows semver)
 // keep in sync with docs/api-description.json
-const SELFOSS_API_VERSION = '4.0.0';
+const SELFOSS_API_VERSION = '5.0.0';
```
davidoskky, line 10: Is the api version 4.1.0 or 5.0.0?

jtojnar: `5.0.0` since the introduction of `403 Forbidden` errors is technically a breaking change.

PR #1382, opened 2022-10-15
Update to react-router 7

jtojnar:
~~Unfortunately, regular expressions in routes are no longer supported so URLs like `/newest/source-2/143` or `/unread/tag-news/259` will no longer work. We will need to switch them to something like `/sources/2?filter=newest#143` and that will collide with the server API, so it will need to be moved to `/api/` subdirectory.~~ I managed to work around https://github.com/remix-run/react-router/issues/8254 with intermediary `EntriesFilter` component with a custom validation logic.

Though I do not really like the direction of the library towards frameworkization.

### DIFF
--- a/client/js/helpers/hooks.js
+++ b/client/js/helpers/hooks.js
@@ -1,5 +1,5 @@
 import { useEffect, useState } from 'react';
-import { useLocation } from 'react-router-dom';
+import { useLocation } from 'react-router';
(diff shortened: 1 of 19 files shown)


jtojnar:
Dropped 18e99da9c003d22b45843b93ead3d879a1ea84d4 in favour of 029abf51ba20364e08c0f29923c506920fc9742f.

PR #1421, opened 2023-03-11
Add 'Share on Mastodon' functionality

dubbl:
Given the recent influx of users on Mastodon, I thought it would be nice to add a "Share on Mastodon" button.

As users are on different mastodon instances with different URLs, we need a new Configuration option to store the URL of the mastodon instance.
I added the option description to the documentation and the English and German translation for the label.

image

The link leads to a page on which the user can edit and publish the mastodon post, pre-filled with the entries title in quotes and its URL.

What do you think?


jtojnar:
Very nice, thank you.

PR #1423, opened 2023-03-28
Improve filter support

jtojnar:
The filter expression can take one of the following forms:

1. An *atomic filter*: a Perl-compatible regular expression, as [accepted by PHP](https://www.php.net/manual/en/reference.pcre.pattern.syntax.php), between forward slashes (`/`). For example, `/reg(ular expression|ex)/` will keep only items whose title or content matches the regular expression between the slashes (i.e. contain the phrase “regular expression” or “regex”). Learn more about regular expressions on <https://www.regular-expressions.info/>.
2. A *field-specific filter*: an *atomic filter* preceded by one of the field names below and a colon:
    - `title:/regex/` will keep only items whose title matches the regular expression between the slashes.
    - `content:/regex/` will keep only items whose content matches the regular expression between the slashes.
    - `author:/regex/` will keep only items which have an author that matches the regular expression between the slashes.
    - `category:/regex/` will keep only items which have a category that matches the regular expression between the slashes.
3. A *negated filter* is either an *atomic filter* or a *field-specific filter* preceded by an exclamation mark (`!`). It will only keep items would not be kept by the filter after the exclamation mark.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,7 +8,7 @@
 - YouTube spout now includes the video description. ([#1412](https://github.com/fossar/selfoss/pull/1412))
 - Mastodon share button was added. Can be enabled by adding `m` to `share` and setting `mastodon` pointing to your chosen instance. ([#1421](https://github.com/fossar/selfoss/pull/1421))
 - Source filters can be negated, or limited to only title or only content. ([#1423](https://github.com/fossar/selfoss/pull/1423))
-- Sources can be filtered based on item’s author. ([#1423](https://github.com/fossar/selfoss/pull/1423))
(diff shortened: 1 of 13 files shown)


GutHib:
Wow, that is truely amazing. Does the project accept donations or should that go towards a charity?

jtojnar:
I can send you an IBAN number if you want to do a bank account transfer. But donating to [Internet Archive](https://archive.org/donate) or [Wikimedia Foundation](https://donate.wikimedia.org) is also fine.

GutHib:
Mail sent to your gmail address, you can find it in your spam folder.

jtojnar:
@GutHib Thank you for your donation :heart:

PR #1433, opened 2023-05-29
feat(docker): add basic docker support

desaintmartin:
Here is a first try of https://github.com/fossar/selfoss/issues/1350 as a simpler alternative / first step than https://github.com/fossar/selfoss/pull/1170

Here are some notes:
- Scope is much smaller than #1170: only having a production container image
- It's a two-stages image, instead of the separated "composer" stage of #1170. ~~If you think this could easily be done, feel free to point me to the right direction, but I was not able to properly separate build from serve (due to always needing to build extensions...)~~ edit: https://hub.docker.com/_/composer/ recommends doing what is done in this pr, so everything seems ok, and I now delete development packages.
- cron in included within the container, which seems to be a very good compromise but may not be ideal in some rare situations
- I'm assuming that SQLite is used and a data volume is mounted and config can be customized (mounted as volume). I haven't tried with external database.
- It's not using fpm but ol' good apache. I'm assuming this is ok for now.

What do you think?

Potential next steps if this gets merged:
- image building & hosting automation to GitHub Container Registry using Github Actions (I can do)
- Kubernetes compatibility through Helm Chart with dedicated database (I can do)
- dev-container
- profit and glory

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,50 @@
+documentation
+.dockerignore
+Dockerfile
+.git
+.git*
+*.md
+
+# compiled output
+/dist
+/tmp
+/out-tsc
+# Only exists if Bazel was run
+/bazel-out
+
+# dependencies
+/node_modules
+
+# profiling files
+chrome-profiler-events.json
+speed-measure-plugin.json
+
+# IDEs and editors
+/.idea
+.project
+.classpath
+.c9/
+*.launch
+.settings/
+*.sublime-workspace
+
+# IDE - VSCode
+.vscode/*
+.history/*
+
+# misc
+/.sass-cache
+/connect.lock
+/coverage
+/libpeerconnection.log
+npm-debug.log
+yarn-error.log
+testem.log
+/typings
+
+# System Files
+.DS_Store
+Thumbs.db
+
+# testing
+junit.xml
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,50 +1,28 @@
-documentation
+# Taken from .gitignore
+/assets/.parcel-cache
+/data/favicons/*.png
+/data/favicons/*.jpg
+/data/thumbnails/*.png
+/data/thumbnails/*.jpg
+/data/cache/*.spc
+/data/logs/*.log
+/data/sqlite/*.db
+/public
+docs/public/
+docs/static/processed_images/
+user.css
+user.js
+*.ini
+node_modules
+.env
+vendor/
+.php_cs.cache
+__pycache__
+
+# Regular docker ignore
 .dockerignore
 Dockerfile
 .git
-.git*
+.github
+.gitignore
 *.md
-
-# compiled output
-/dist
-/tmp
-/out-tsc
-# Only exists if Bazel was run
-/bazel-out
-
-# dependencies
-/node_modules
-
-# profiling files
-chrome-profiler-events.json
-speed-measure-plugin.json
-
-# IDEs and editors
-/.idea
-.project
-.classpath
-.c9/
-*.launch
-.settings/
-*.sublime-workspace
-
-# IDE - VSCode
-.vscode/*
-.history/*
-
-# misc
-/.sass-cache
-/connect.lock
-/coverage
-/libpeerconnection.log
-npm-debug.log
-yarn-error.log
-testem.log
-/typings
-
-# System Files
-.DS_Store
-Thumbs.db
-
-# testing
-junit.xml
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,55 @@
+### Stage 1: build client
+FROM node:18 as client-builder
+WORKDIR /client-builder
+
+# Install node packages
+COPY package.json .
+COPY client/package.json client/
+COPY client/package-lock.json client/
+RUN npm run install-dependencies:client
+
+# Build client
+COPY client/ client/
+RUN npm run build
+
+
+### Stage 2: final container
+FROM php:8.2-apache
+RUN apt-get update \
+    && apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install --no-install-recommends -y cron unzip libjpeg62-turbo-dev libpng-dev libpq-dev libonig-dev libtidy-dev \
+    && update-ca-certificates --fresh \
+    && apt clean \
+    && rm -rf /var/lib/apt/lists/*
+
+RUN docker-php-ext-configure gd \
+    && docker-php-ext-install gd mbstring pdo_pgsql pdo_mysql tidy
+
+RUN a2enmod headers rewrite
+
+RUN php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');" \
+    && php composer-setup.php \
+    && php -r "unlink('composer-setup.php');" \
+    && mv composer.phar /usr/local/bin/composer
+
+# Install dependencies
+COPY composer.json .
+COPY composer.lock .
+RUN COMPOSER_ALLOW_SUPERUSER=1 composer install --optimize-autoloader --no-dev
+
+# Setup cron
+RUN echo '* * * * * curl http://localhost/update' | tee /etc/cron.d/selfoss \
+    && chmod 0644 /etc/cron.d/selfoss \
+    && crontab /etc/cron.d/selfoss
+
+WORKDIR /var/www/html
+
+COPY . .
+
+COPY --from=client-builder /client-builder/public /var/www/html/public
+
+RUN chown -R www-data:www-data /var/www/html/data
+
+# Overload default command to run cron in the background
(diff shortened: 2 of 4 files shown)


.dockerignore
jtojnar, line 1: I do not really like these huge blacklists with patterns that are not relevant. Maybe just copy what we need. Or better copy the result on `npm run dist`, which also prunes the `vendor/` directory.

Dockerfile
jtojnar, line 1: IIUC, each `RUN` creates a new layer which is suboptimal https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#minimize-the-number-of-layers

Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+    && php -r "unlink('composer-setup.php');" \
+    && mv composer.phar /usr/local/bin/composer
+
+# Install dependencies
+COPY composer.json .
+COPY composer.lock .
+RUN COMPOSER_ALLOW_SUPERUSER=1 composer install --optimize-autoloader --no-dev
+
+# Setup cron
+RUN echo '* * * * * curl http://localhost/update' | tee /etc/cron.d/selfoss \
```
jtojnar, line 40: Does cron run multiple instances or wait for the old one to finish? If the former, that sounds like a recipe for #967. Either way running every minute is probably too frequent.

Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+WORKDIR /var/www/html
+
+COPY . .
+
+COPY --from=client-builder /client-builder/public /var/www/html/public
+
+RUN chown -R www-data:www-data /var/www/html/data
+
+# Overload default command to run cron in the background
+RUN sed -i 's/^exec /service cron start\n\nexec /' /usr/local/bin/apache2-foreground
```
jtojnar, line 53: This looks too messy for my taste. At minimum we should use a service manager like https://gitlab.com/radek-sprta/docker-selfoss does.

Ideally, the whole thing would be fully declarative like #1271 but I still need to figure out how to use s6-rc service manager.

jtojnar (commented):
Thanks for the contribution.

Still the main obstacle is that I do not understand how users that want Docker expect to use the image so I have no way of evaluating whether the PR fits that.

Do people just run `docker` commands manually? Or use something like docker-compose? How do they handle updates? What are security considerations? How do people configure software? How do they manage data backups?

desaintmartin:
A few months later, let me try to answer.

> Still the main obstacle is that I do not understand how users that want Docker expect to use the image so I have no way of evaluating whether the PR fits that.

I understand this issue. Maybe I can just say that I'm a Kubernetes contributor (see my profile) and my production clusters manage ~15000 running containers right now, if that helps. Of course I don't want to pretend to be someone here! ;)

> Do people just run `docker` commands manually? Or use something like docker-compose? How do they handle updates? What are security considerations? How do people configure software? How do they manage data backups?

It really depend, some use docker-compose, some use raw docker/containerd commands, some use Kubernetes either directly or through Helm, some use home NAS like synology to manage containers for them, but all of them use container image definition.

I'm going to update a bit the PR to be more up-to-date and take into account your suggestions!

Dockerfile
desaintmartin: Actually, the more I think about it, the more I think maybe we should NOT try to deal with it from inside of the container. A container is supposed to be simple and run one thing, and put cron AND apache on the same container is more of a hack. At worse, one still have to configure its own cron, at best we can automate further through another container in the future.

What's your opinion?

desaintmartin:
Notes:
- I haven't included a docker healthcheck. I am not sure it is always wanted. On kubernetes, healthcheck are a completely different system anyway
- I disabled cron, so we still have to do it like today. Much more atomic / simpler like a container is supposed to be, much less surprise / bugs.
- `php:8.2-apache` image, built on top of `debian:12-slim`, is 710MB (!!), final image is 770MB so 60MB more than upstream image, which is not bad, relatively speaking (absolutely speaking, this is huge, but this seems to be how it works today).
- We need minimalistic documentation, just tell me where you want it to be.

Dockerfile
dxlr8r: could have another container doing that job. For Kubernetes, you could simply use a CronSet.

desaintmartin:
Hello @jtojnar, would you have any time to review this PR?

desaintmartin:
I would really love to use selfoss from a container so that I can automate / simplify my self-hosting, and at the same time I would hate to have a long-running fork of selfoss. What is your opinion on having an official image @jtojnar ?

jtojnar:
Sorry about not replying sooner, I have opened the tab multiple times but then got distracted each time.

I think this is valuable but at the same time I still worry about my ability to keep it working, as I do not use Docker. Especially when I have had less time for my open-source projects since I started a full-time job.

After giving it some more thought, I think I would be fine with merging this as long as the level of support is clearly communicated. For example, we could have it called `selfoss-community/selfoss` in Docker Hub. We should also mention in the description who is currently maintaining it (even explicitly inviting people to try to increase the bus factor).

I could also grant you and @radek-sprta commit access to this repo so you could maintain this without depending on me for merging.

Alternately, we could create a separate `selfoss-docker` repo. This would allow us to invite even more people as committers without having to trust them with the main repo. But that would probably require annoying amounts CI code for synchronization.

dependabot.yml
jtojnar, line 1: Could you move this into the `.github` subdirectory, as to not clutter the root?

package.json
```diff
@@ -32,6 +32,7 @@
     "fix:helpers:cs": "black utils/ tests/",
     "install-dependencies": "npm run install-dependencies:client && npm run install-dependencies:server",
     "install-dependencies:client": "npm install --production=false --prefix client/",
+    "install-dependencies-ci:client": "npm ci --production=false --prefix client/",
```
jtojnar, line 35: I think it would make sense to just use `npm ci` for `install-dependencies:client` – reading the [docs](https://docs.npmjs.com/cli/v10/commands/npm-ci), I do not see why one would want an unclean installation of dependencies.

desaintmartin:
Either ways are fine for me, and of course I think, as there seem to be several people interested into maintaining a Docker image with several tries over the years, that we should find a way to have the most maintainable solution.

Pros of having a separate repo: more secure

Cons of having a separate repo: 
- Usual way to build a container is from sources, so it would actually... be a fork?
- we can't use the github container registry in a transparent manner with UI from the main repository (See https://github.com/desaintmartin/selfoss for an example, for now with unpublished versions). But that may be against your goals of marking it as "unofficial".

I don't think too many people with write access (btw, not push, only merge!) are needed.

.dockerignore
jtojnar: Right. Which is why I also suggested using `npm run dist` command. It will build a production package, including [filtering the PHP directories](https://github.com/fossar/selfoss/blob/f75240b200f8d56478d245fcb8e648229ccd241d/utils/create-zipball.py). It would allow us to clean the Dockerfile even further and we would not need the `.dockerignore`.

Of course, it would mean building the image from scratch every time but perhaps that is fine for production/non-dev image.

jtojnar:
> Usual way to build a container is from sources, so it would actually... be a fork?

Not necessarily, see e.g. https://github.com/wallabag/docker. That one currently only builds tagged versions but nighlties could be handled by GitHub actions in the main repo [triggering](https://blog.marcnuri.com/triggering-github-actions-across-different-repositories) builds in the docker one.

> * we can't use the github container registry in a transparent manner with UI from the main repository

Yeah, I think that would be a good thing – at least until selfoss development team contains enough people who could maintain it so that the risk of it becoming unmaintained is reduced.

> Cons of having a separate repo:

Additionally, separating would make synchronous changes more work: contributor would need to make PRs in two different places. And the repos could get out of sync breaking compatibility.

That could be minimized by using a single entry point like `nix run dist` and keeping the structure unchanged. But I think I still lean towards using the main repository for this reason.

> I don't think too many people with write access (btw, not push, only merge!) are needed.

Right, we could require going through PRs and then adding review requirement in branch protections.

dependabot.yml
desaintmartin: It was mostly a test from me, I've removed it as it may change behavior of other, not docker-related, dependencies upgrades!

package.json
desaintmartin: Indeed, fixed!

.dockerignore
desaintmartin: Actually, when you `COPY` something into a container image, it can't get removed, so if we copy local data, for example, for local builds, it will go into the final image size even if scripts inside container build filter them. This is a good practice in the docker community to filter out everything that could exist in a local development environment to avoid side effects.

nijave:
Made a packaging repo where `renovate` should auto merge PRs with git submodule updates (this repo) and create new builds https://github.com/nijave/Selfoss-Docker

Based off @matthk

PR #1440, opened 2023-06-15
Add a selfoss CLI tool

jtojnar:
This replaces the `cliupdate.php` script with a `symfony/console` based entry point:

    php bin/selfoss update

In the future we are going to introduce more commands.


jtojnar:
I do not really like Symfony console.

`ContentLoader` used transitively by `UpdateCommand` is obtaining a `DatabaseConnection` through dependency injection. As a result, every time `selfossctl` is run, it will attempt to connect to database.

It is possible to [make loading the commands lazy](https://symfony.com/doc/current/console/lazy_commands.html) using `ContainerCommandLoader` (if it were not [buggy with slince/di](https://github.com/slince/di/issues/10)) but because that does not have access to command descriptions, it will still require instantiating all commands when doing `selfossctl list`.

Now, Symfony Console also supports [lazy command descriptions](https://symfony.com/blog/new-in-symfony-5-3-lazy-command-description) but I would either have to use Symfony’s DI container (regular commands are getting turned to lazy ones in [`AddConsoleCommandPass`](https://github.com/symfony/symfony/blob/52a92926f7fed15cdff399c6921100a10e0d6f61/src/Symfony/Component/Console/DependencyInjection/AddConsoleCommandPass.php#L119-L120)) or create `LazyCommand`s manually (that’s what we are doing now).

And even then `selfossctl help update` still requires instantiating the command.

PR #1460, opened 2023-07-29
client: Restore “Mark all” button to 2.18 behaviour

jtojnar:
In 2.18, when in “unread” mode, if there were no unread items, clicking the button would clear out all items and trigger a reload or refresh status of the “Load more” button, depending on whether the client was aware of more unread items on the server.

When we ported entries to React in f3b279c808e8a5d7e3ab5c8254bdda83c094b50e, we accidentally broke the “Mark all as read” button, making it keep all the items it marked in the view (a). We also did not port the code that would load more items afterwards (b).

It got further broken during porting routing to React in 405a3ec2cf49e3dc17421b997930c5b1faebd2fc. We looked for the `filter` field on a wrong object (c).

We fixed (c) and attempted to fix (a) in ec544b7ae066050dcdb43e362f38854d4fdb7bea but the patch just made it keep all the items it did not mark, such as those marked as read manually beforehand, instead.

This patch makes it so that the button removes all the items in the view except for the active item and loads more items afterwards, returning to behaviour similar to 2.18.


jtojnar:
Because the `entries` are now cleared, we cannot use the simple hack of just clicking the “Load more” button but need a proper dispatcher to be able to pass the `entries` before deletion.

Unfortunately, the dispatcher appears to bring another bug:

0. Set `auto_stream_more=0` and `items_perpage=2`
1. In “Unread” mode click “Mark as read” button.
2. Click “Newest” mode.
3. Observe that the “Load more” spinner will be stuck on spinning.

jtojnar:
I also noticed the following weird behaviour but that does not appear to be a regression:

1. In  “Newest”, click “Load more”.
2. Click  “Newest” again.
3. Last loaded page will be appended again.

PR #1469, opened 2023-11-23
client: Double click on title to mark entry as read

PhrozenByte:
Ten years after #450 (I'm feeling old...) it's time to try again to finally bring feature to upstream: Double click on an entry's title to mark it as read (and double click again to mark it as unread; disabled on smartphones due to the different UI concept).

I believe that this should be the default behaviour, but since it must add a delay to distinguish between single and double clicks, some people might find it obstructive, thus I've deliberately added it as opt-in using the new `double_click_mark_as_read` option.

This was the first time I worked with React. So please check whether the code I've written (including the `handleMultiClick` function and its usage, I'm not sure whether `useCallback` still makes sense there) is like it's expected for React.

### DIFF
--- a/client/js/templates/Item.jsx
+++ b/client/js/templates/Item.jsx
@@ -257,6 +257,8 @@ export default function Item({ currentTime, item, selected, expanded, setNavExpa
         [currentTime, item.datetime]
     );
 
+    const canWrite = useAllowedToWrite();
+
     const previouslyExpanded = usePreviousImmediate(expanded);
     const configuration = useContext(ConfigurationContext);
 
@@ -410,8 +412,6 @@ export default function Item({ currentTime, item, selected, expanded, setNavExpa
         }),
         [item.source]
     );
-
-    const canWrite = useAllowedToWrite();
 
     const _ = useContext(LocalizationContext);
 
--- a/client/js/templates/Item.jsx
+++ b/client/js/templates/Item.jsx
@@ -46,6 +46,30 @@ function setupLightbox({
     }));
 }
 
+function useMultiClickHandler(handler, delay = 400) {
+    const [state, setState] = useState({ clicks: 0, args: [] });
+
+    useEffect(() => {
+        const timer = setTimeout(() => {
+            setState({ clicks: 0, args: [] });
+
+            if (state.clicks > 0 && typeof handler[state.clicks] === 'function') {
+                handler[state.clicks](...state.args);
+            }
+        }, delay);
+
+        return () => clearTimeout(timer);
+    }, [handler, delay, state.clicks, state.args]);
+
+    return (...args) => {
+        setState((prevState) => ({ clicks: prevState.clicks + 1, args }));
+
+        if (typeof handler[0] === 'function') {
+            handler[0](...args);
+        }
+    };
+}
+
 function stopPropagation(event) {
     event.stopPropagation();
 }
@@ -104,7 +128,7 @@ function closeFullScreen({ event, history, location, entryId }) {
 }
 
 // show/hide entry
-function handleClick({ event, history, location, expanded, id, target }) {
+function handleToggleOpenClick({ event, history, location, expanded, id, target }) {
     const expected = selfoss.isMobile() ? '.entry' : '.entry-title';
     if (target !== expected) {
         return;
@@ -121,6 +145,14 @@ function handleClick({ event, history, location, expanded, id, target }) {
         selfoss.entriesPage.activateEntry(id);
         history.replace(makeEntriesLink(location, { id }));
     }
+}
+
+// mark entry read/unread
+function handleToggleReadClick({ event, unread, id }) {
+    event.preventDefault();
+    event.stopPropagation();
+
+    selfoss.entriesPage.markEntryRead(id, unread == 1);
 }
 
 // load images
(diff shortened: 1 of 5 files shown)


client/styles/main.scss
```diff
@@ -523,6 +523,7 @@ span.offline-count.diff {
     color: #999999;
     padding-top: 7px;
     padding-bottom: 7px;
+    user-select: none;
```
jtojnar, line 526: This will make it impossible to select the title to copy/search it.

client/js/templates/Item.jsx
```diff
@@ -444,7 +487,7 @@ export default function Item({ currentTime, item, selected, expanded, setNavExpa
             {/* title */}
             <h3
                 className="entry-title"
-                onClick={titleOnClick}
+                onClick={configuration.doubleClickMarkAsRead ? titleOnMultiClick : titleOnClick}
```
jtojnar, line 490: We would probably want to skip smartphones here to avoid the delay there.

client/js/templates/Item.jsx
```diff
@@ -360,15 +394,30 @@ export default function Item({ currentTime, item, selected, expanded, setNavExpa
...
+        (event) => handleToggleOpenClick({ event, history, location, expanded, id: item.id, target: '.entry-title' }),
         [history, location, expanded, item.id]
     );
 
+    const titleOnMultiClick = useMultiClickHandler({
+        0: (event) => {
+            event.preventDefault();
+        },
+        1: titleOnClick,
+        2: useCallback(
```
jtojnar, line 411: `useCallback` would be needed if `handler[2]` was used in dependencies list of `useEffect` but in this case, the whole `handler` object is used so that needs to be wrapped in `useMemo` instead, see https://react.dev/reference/react/useMemo#memoizing-a-dependency-of-another-hook

client/js/templates/Item.jsx
jtojnar, line 1: I do not like the delay. selfoss is already far too slow for my taste (still have some optimizations in the pipeline) so it cannot be the default.

The code is pretty nice and I can see how easier access to toggling read status without opening the entry would be convenient but I am still not sure double click is the right action for it. Usually double click action is used for opening e.g. file, contrasted with single click for selection. That also works without introducing delay because selection can be immediate.

Personally, I use `auto_mark_as_read=1` which makes marking articles as read convenient. Though it will not help with unmarking. Maybe we could also consider other alternatives like point 3. from https://github.com/fossar/selfoss/issues/969#issuecomment-323085061

docs/content/docs/administration/options.md
```diff
@@ -277,6 +277,12 @@ Hide read articles on mobile devices.
 Set to `0` to stop the interface from scrolling to the article header when an article is opened.
 </div>
 
+### `double_click_mark_as_read`
+<div class="config-option">
+
+set this to `1` to mark an item as read when double clicking on it.
```
jtojnar, line 283:
```suggestion
Set this to `1` to enable toggling item’ (un)read state by double clicking the title.
```

client/js/templates/Item.jsx
PhrozenByte: Agreed.

client/js/templates/Item.jsx
PhrozenByte: True. How about dumping the delay?

It's just important to note that a double click then fires both events, i.e. a double click causes the entry to expand (or collapse) and marks it as read (or unread). It's a little less separation of concerns, but still quite convenient: You click once to expand the article, read it and close+mark it with a double click. If you decide not to read it yet (that's why I don't like `auto_mark_as_read=1`: I often decide not to read an article straight away after expanding it, but later), you just collapse it with a single click. The only disadvantage is that one can't mark an entry as read without opening it. That's where https://github.com/fossar/selfoss/issues/969#issuecomment-323085061 might come to a rescue: If we also add the shortcut icons, we can have it all.

client/styles/main.scss
PhrozenByte: That's intentional, because double clicks also mark text. However, thinking about it, why isn't `event.preventDefault()` preventing that? I'll look into this whether there's a better solution, but first let us decide what to do with the delay (see below).

client/js/templates/Item.jsx
PhrozenByte: I probably miss a lot of React's basic rendering concept to truly understand that :see_no_evil: So, please excuse if I don't understand it right away, but from your info and the docs, combined with my limited understanding, it should look like the following? :thinking: 

```javascript
    const titleOnMultiClick = useMultiClickHandler(useMemo(() => {
        0: (event) => {
            event.preventDefault();
        },
        1: (event) => handleToggleOpenClick({ event, history, location, expanded, id: item.id, target: '.entry-title' }),
        2: (event) => {
            if (canWrite && !selfoss.isSmartphone()) {
                handleToggleReadClick({ event, unread: item.unread, id: item.id });
            }
        )
    }, [ canWrite, history, location, expanded, item.unread, item.id ]));
```

client/styles/main.scss
jtojnar: Yes, I know, I usually do double click to select words.

client/js/templates/Item.jsx
jtojnar: Yes, that looks correct to me.

client/js/templates/Item.jsx
PhrozenByte: @jtojnar Can you please take a look into this? Just need a quick decision, I'll implement it accordingly then.

Issue #1411, opened 2023-02-12
Open Link instead of expand the Item?


spec1re:
Is it possible to somehow just open the the link instead of expanding it, like clicking on the favicon or datetime?

jtojnar:
There is currently no configuration option. You could probably add some code to `user.js` (I can come up with something later if you want it).

It might be interesting to implement some kind of event-action system that would allow remapping the `item:activate` action to `open-link`.

spec1re:
> There is currently no configuration option. You could probably add some code to `user.js` for now. (I can come up with something later if you want.)
> 
> It might be interesting to implement some kind of event-action system that would allow remapping the `item:activate` action to `open-link`.

That would be awesome! Thanks.

I really like Selfoss, I would like to see a "light" mode, no content just links to the news. This would be more consistent, just open a new tab.

spec1re:
How do I do it with user.js, sorry zero JS knowledge here. 

Thanks.

jtojnar:
This is ugly but seems to work:

```javascript
function openEntryLink(id) {
    window.open(document.querySelector(`.entry[data-entry-id="${id}"]`).getAttribute('data-entry-url'));
}

const entriesPageNotifier = {
    set: function(obj, prop, value) {
        console.log(obj, prop, value)
        obj[prop] = value;

        if (prop === 'entriesPage' && value !== null) {
            value.activateEntry = openEntryLink;
        }

        return true;
    }
};

selfoss = new Proxy(selfoss, entriesPageNotifier);

if (selfoss.entriesPage !== null && selfoss.entriesPage.activateEntry !== openEntryLink) {
    selfoss.entriesPage.activateEntry = openEntryLink;
}
```

spec1re:
@jtojnar 

No this is beautiful and fully serves the purpose. 

Thank you very much, much appreciated!

spec1re:
```
function openEntryLink(id) {
    window.open(document.querySelector(`.entry[data-entry-id="${id}"]`).getAttribute('data-entry-url'));
	document.getElementById("currentpage").innerHTML = window.location.replace;
}

const entriesPageNotifier = {
    set: function(obj, prop, value) {
        console.log(obj, prop, value)
        obj[prop] = value;

        if (prop === 'entriesPage' && value !== null) {
            value.activateEntry = openEntryLink;
        }
        return true;
    }
};

selfoss = new Proxy(selfoss, entriesPageNotifier);

[...]

PR #1471, opened 2023-11-23
client: Middle click on title to open entry in new tab

PhrozenByte:
Ten years after #452 (I'm feeling old...) it's time to try again to finally bring this tech-wise tiny but UX-wise major feature to upstream. I'm not aware of any good reason why one should be required to click on the tiny favicon to open an entry in a new tab.

I believe that this should be default behaviour. There's still the possibility to make this an opt-in feature using a new `config.ini` option, but I'd strongly discourage it - it's simple and basic, it doesn't affect other users, thus it shouldn't come with unnecessary complexity (like an option).

If you want to merge both # and #, ignore this PR and merge # instead. Merging any of the three PRs will automatically close the other two PRs.

Closes #
Closes #

### DIFF
--- a/client/js/templates/Item.jsx
+++ b/client/js/templates/Item.jsx
@@ -446,11 +446,11 @@ export default function Item({ currentTime, item, selected, expanded, setNavExpa
                 className="entry-title"
                 onClick={titleOnClick}
             >
-                <span
+                <a
+                    href={item.link}
                     className="entry-title-link"
                     aria-expanded={expanded}
                     aria-current={selected}
-                    role="link"
                     tabIndex="0"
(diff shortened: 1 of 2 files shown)


PhrozenByte:
Related issue: #1411

jtojnar:
I think this makes sense. Using a link might be a bit iffy from accessibility standpoint but then again, we used ARIA `role=link` before so this should not be a regression.

jtojnar:
Actually, now I remember why we do not do this. Some spouts (e.g. Twitter) contain links in the title and links cannot be nested. I guess we can fall back to the old way when the title contains `<` character.

PR #1574, opened 2026-07-26
spouts/rss/enclosures: Link enclosed audio file

jtojnar:
This is useful for playing podcasts.

Not embedding audio player for now, as that would require loading on demand to prevent tracking.

See: https://github.com/fossar/selfoss/issues/1169


otyugh:
Being impatient made me do a custom Fossar instance :p (I didn't even hope you'd answer, I'm baaad)

I just did this to my selfoss/src/spouts/rss/feed.php, assuming that, if there is an enclosure, that's what we want to load (work for most podcast).
```
            $enclosure=$item->get_enclosure()->link;
            if(!empty($enclosure)){$link=$enclosure;}else{$link = htmlspecialchars_decode((string) $item->get_link(), ENT_COMPAT);}
```

PR #1610, opened 2026-08-16
docs: Use shortcuts from json file

jtojnar:


docs/static/shortcuts.json
jtojnar, line 1: I would prefer to put this in `client/js` directory, not `docs` but then zola would complain that it is “not inside the base site directory”, even when using a symlink.

Denperidge: Does it work in reverse? Symlink the zola file towards client/js?

jtojnar: Yes, that should work. But client can just import the json file directly.

